PR #35, opened 2016-01-10
Fix XSS in all messages using the pod's name.

TheNeikos:
This commit fixes a Persistent XSS. The problem was that at no point the output
was sanitized, allowing each pod to control the column 'name' in the database.
Since this can be set to anything one wants it can be malicious.
Using htmlentities(name, ENT_QUOTES) should be sufficient to ward this off.

------

I have found about this due to my pod's name having an apostrophe, cutting off text at first. A quick check in the HTML confirmed my fears, and a check in the codebase found the source. 

This PR should fix *that* instance of the problem, but looking at the codebase I can see a deeper problem as it does not seem maintainable. But I digress.

This has not been tested extensively, so please do scrutinize and test it locally before merging, I would not want to create additional security issues.

### DIFF
--- a/api.php
+++ b/api.php
@@ -22,8 +22,9 @@ EOF;
  }
  $numrows = pg_num_rows($result);
 while ($row = pg_fetch_array($result)) {
+$pod_name = htmlentities($row["name"], ENT_QUOTES);
 $tip="";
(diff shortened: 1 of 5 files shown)


dmorley:
Thanks. Of course Github did not notify me of this pr. I think I have made changes since, but will check it out!

TheNeikos:
Okay, I fixed the merge conflict.

dmorley:
Great. Thank you! I think I have you as a contact in d*?

TheNeikos:
dmorley, yup! I added you, but I didn't manage to find out how to message you, and still don't
If you want you can send me an e-mail (neikos at neikos.email) so we can get into contact

PR #38, opened 2016-06-02
JSON output for api-more.php

gsantner:
* JSON (via GET Parameter) output for api-more.php
* local_posts and comment_counts added to api.php (= Also maybe an important information for choosing an pod)


gsantner:
Project seems dead... or at least no active maintainers....

dmorley:
Sorry I missed this. I added manually these things and fixed the dead api

PR #41, opened 2016-12-21
Improve config.php file inclusion

noplanman:
Move config.php to the root level, as it doesn't only have the DB connection info.
Instead of just including the config, it must be required, as it is an essential file.
Move require statement below $_GET checks.


dmorley:
Nice, thx

PR #44, opened 2016-12-21
Remove stray marker in the Gulf of Guinea.

noplanman:
Remove the empty stray marker from the map.
Fix the comma separation of markers.

stray-marker


dmorley:
How do you know no pod is on a barge there! ok ok

noplanman:
> How do you know no pod is on a barge there!

Awesome! A pod that could move into international waters! 😄

PR #51, opened 2016-12-22
Clean up most of the PHP code.

noplanman:
Big sweep through the PHP code, making it more robust.

### DIFF
--- a/api.php
+++ b/api.php
@@ -7,12 +7,10 @@ require_once __DIR__ . '/config.php';
 $dbh = pg_connect("dbname=$pgdb user=$pguser password=$pgpass");
 $dbh || die('Error in connection: ' . pg_last_error());
 
-
-if ($_GET['format'] == 'georss') {
+if ($_GET['format'] === 'georss') {
   echo <<<EOF
 <?xml version="1.0" encoding="utf-8"?>
-<feed xmlns="http://www.w3.org/2005/Atom"
-xmlns:georss="http://www.georss.org/georss">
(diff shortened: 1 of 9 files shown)


dmorley:
So this seems to be a magical if, then statement very cool. Just trying to understand why only used here and not all if, then statements? Because it is just a simple if this then print that type thing?

$row['service_facebook'] === 't' && print '<div class="smlogo smlogo-facebook"></div>';

noplanman:
What other places are you talking about?

I tend to opt for the simple one-liners for readability, like for the services where there are 5 (almost) identical lines.

dmorley:
But it could be used for something like: 

 if ($_GET['format'] === 'json') {
     echo json_encode($row);
   } else {

if you wanted?

noplanman:
True, but what about the else part?

For some situations, things can be reused and bent, like this for example:
```php
if ($_GET['method'] === 'jsonp') {
  print $_GET['callback'] . '(' . json_encode($obj) . ')';
} else {
  print json_encode($obj);
}
```
would become (using the ternary operator):
```php
$obj_json = json_encode($obj);
echo $_GET['method'] === 'jsonp' ? $_GET['callback'] . '(' . $obj_json . ')' : $obj_json; 
```

But looking at readability, the second is quite difficult to understand.

dmorley:
yea, I am lost looking at that, it's pretty and all, but my brain works better with if, then, do or else type things. I learned Logo then Basic in the 80s, still on that level!

noplanman:
Readability is key!
If it's not readable and understandable, it's bad code, at least to me.

Main thing is for you to understand the code, so if there are any bits that seem weird or illogical, just comment on that line and let me know!

Issue #45, opened 2016-12-21
Simplify database


noplanman:
At the moment, the database has many columns that should be altered.

**Wrong Types**
e.g. these should all just be boolean:
`hidden`, `ipv6` (string "yes" or "no")
`secure` (string "true" or "false")
`signup` (integer 1 or *null*)
`service_facebook` etc. (are boolean)

**Rename**
New column `xmpp` should be renamed to `service_xmpp` to follow the naming convention of the other services.
Integer columns that are summed up numbers should be named similarly (`total_*` vs. `*_count` vs. just a name like `local_posts`)

**Redundant Columns**
`connection`, `longversion`, `shortversion` and `masterversion` all have similar values.
This could be simplified to values that make perfect sense.
What's the difference between `shortversion` and `masterversion`? Some pods have totally different values here.

dmorley:
Agree it makes sense to change the types and rename. 
shortversion=6.2.0 (basic version pulled from the pod)
longversion=6.2.0-head059840598 (I think no one cares, this could be dropped, it's not a human readable version)
masterversion=6.2.0 (this is taken from the github repo, what a up to date pod should be running based on if it is diaspora, or frendica etc), this should really be its own table of software versions and current revisions.

dmorley:
also pingdom was the original provider for stats I used and they are no longer free so I do not think any pods are using their paid service. It would make sense to change the 'pingdomurl' and more to generic 'statsurl' etc. The Hxxx ones are also from before when pods did not have a json output and I scraped headers, can go away. 

What is the best way to clean this all up? In David land I would edit the tables.sql and the code that needs changing and then go and do the sql updates manually, however migration would make more sense of course. Ill poke around on the interwebs, but no idea what is cool these days, probably something.

Also /db/ folder should just be db items I think and all the php scripts should be in main dir? Who made this mess?

noplanman:
I'd do it like you said. Instead of manually changing the DB structure, I'd write a script that does it for you. Especially as some things need special altering, like making all boolean fields also have boolean values. Not sure if just changing the type automatically does it right.

Also, having a script makes the deployment easier. Just update the new files and run the script, no downtime or weirdness.

noplanman:
Just linking your gist here for easier access:
https://gist.github.com/dmorley/0a2470c91fcc5a6f583c67f51ba409f1

dmorley:
[...]

PR #64, opened 2017-01-11
fix using the whois column for random location data

dmorley:
Stop filling whois with IP deets. Show city, state, country rather than lat, long humans don't use lat long in viewing location of anything.

### DIFF
--- a/db/api-more.php
+++ b/db/api-more.php
@@ -27,7 +27,9 @@ while ($row = pg_fetch_array($result)) {
     echo 'Months Monitored: ' . $row['monthsmonitored'] . '<br>';
     echo 'Response Time: ' . $row['responsetimelast7'] . '<br>';
     echo 'User Rating: ' . $row['userrating'] . '<br>';
-    echo 'Server Location: ' . $row['country'] . '<br>';
+    echo 'Server Country: ' . $row['country'] . '<br>';
+    echo 'Server State: ' . $row['state'] . '<br>';
+    echo 'Server City: ' . $row['city'] . '<br>';
     echo 'Latitude: ' . $row['lat'] . '<br>';
     echo 'Longitude: ' . $row['long'] . '<br>';
   }
--- a/show.php
+++ b/show.php
@@ -56,9 +56,9 @@ $numrows = pg_num_rows($result);
     echo '<td>' . $row['uptimelast7'] . '%</td>';
     echo '<td data-toggle="tooltip" data-placement="bottom" title="active six months: ' . $row['active_users_halfyear'] . ', active one month: ' . $row['active_users_monthly'] . '">' . $row['active_users_halfyear'] . '</td>';
     if ($country_code === $row['country']) {
-      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: ' . $row['city'] . ' State: ' . $row['state'] . '"><b>' . $row['country'] . '</b></td>';
+      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: '; echo $row['city'] ?? 'n/a'; echo ' State: '; echo $row['state'] ?? 'n/a'; echo '"><b>' . $row['country'] . '</b></td>';
     } else {
-      echo '<td data-toggle="tooltip" data-placement="bottom" title="City: ' . $row['city'] . ' State: ' . $row['state'] . '">' . $row['country'] . '</td>';
+      echo '<td data-toggle="tooltip" data-placement="bottom" title="City: '; echo $row['city'] ?? 'n/a'; echo ' State: '; echo $row['state'] ?? 'n/a'; echo '">' . $row['country'] . '</td>';
     }
     echo '<td>';
     $row['service_facebook'] === 't' && print '<div class="smlogo smlogo-facebook" title="Publish to Facebook" alt="Publish to Facebook"></div>';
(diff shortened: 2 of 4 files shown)


noplanman:
I think it would make sense to remove the `whois` column completely, as it just has redundant data in there, combining the `country`, `lat` and `long` columns.

Nudge towards #45...

dmorley:
Yea whois is supposed to be true domain whois I at some point started putting shit in it, this is to back out doing that. I do not think it makes sense to try and pull whois data, so let me add that to the sql plan separately and keep all the sql changes as one.

noplanman:
Ok, that makes sense! Will merge here...

db/api-more.php
```diff
@@ -27,7 +27,9 @@
     echo 'Months Monitored: ' . $row['monthsmonitored'] . '<br>';
     echo 'Response Time: ' . $row['responsetimelast7'] . '<br>';
     echo 'User Rating: ' . $row['userrating'] . '<br>';
-    echo 'Server Location: ' . $row['country'] . '<br>';
+    echo 'Server Country: ' . $row['country'] . '<br>';
+    echo 'Server State: ' . $row['state'] . '<br>';
```
noplanman: I've noticed in the db dump that I have, that all `state` and `city` fields are empty.
That results in a text without any actual data.

Instead, to make it look less like an error, adding something like `n/a` or something would let the user know that it's just not known.

This applies for the other files/changes too.

dmorley: That make sense for here?

noplanman: Here maybe not, but even so it should be a string and not `null` for the echo statement.
For now though, I think it's ok to leave it as is.

But I'd add something on the `show.php` and `showfull.php` files, as the popup might otherwise look like this, which is a bit pointless:
`City: State:`

dmorley: Oh I added it to the pull.php so it puts a n/a in the DB on these fields. My brain went the DB way vs the display way, updated PR

noplanman: Erm, did I miss the update or have you not pushed yet?

dmorley: https://github.com/diasporg/Poduptime/pull/64/commits/cacff7a82dd71066a9628f76197b57080f2a8b38

show.php
```diff
@@ -56,9 +56,9 @@
     echo '<td>' . $row['uptimelast7'] . '%</td>';
     echo '<td data-toggle="tooltip" data-placement="bottom" title="active six months: ' . $row['active_users_halfyear'] . ', active one month: ' . $row['active_users_monthly'] . '">' . $row['active_users_halfyear'] . '</td>';
     if ($country_code === $row['country']) {
-      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="' . $row['whois'] . '"><b>' . $row['country'] . '</b></td>';
+      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: ' . $row['city'] . ' State: ' . $row['state'] . '"><b>' . $row['country'] . '</b></td>';
```
noplanman: I was actually talking about these guys, that get displayed when hovering the country code.
(same for `showfull.php`)

dmorley: I guess my thought it that pull.php should keep improving to fill in the DB so that nothing is just a blank anymore, that way even on the api and other spots it will show city:n/a, so the next run to update your db it will not be blank anymore. Are we talking different things here?

noplanman: Now I see what you mean. In my opinion, it's better to just leave `null` in the database if there is no data, not have a bunch of `n/a` strings in there.
It's up to the code to decide what to do with the data.

If you save `n/a`, there is no way of knowing if there actually is good data or not, as it's a normal string, like the real data would be too!

dmorley: I kind of thought that too just as I sent that comment :) I'll fix now!

dmorley: Is their a more graceful way than:

title="City: '; echo isset($row['city']) ? $row['city'] : "n/a"; echo ' State: ....

looks so clunky

noplanman: Yep, PHP7 way 😉 
`$row['city'] ?? 'n/a'`

dmorley: better?

show.php
```diff
@@ -56,9 +56,9 @@
     echo '<td>' . $row['uptimelast7'] . '%</td>';
     echo '<td data-toggle="tooltip" data-placement="bottom" title="active six months: ' . $row['active_users_halfyear'] . ', active one month: ' . $row['active_users_monthly'] . '">' . $row['active_users_halfyear'] . '</td>';
     if ($country_code === $row['country']) {
-      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: ' . $row['city'] . ' State: ' . $row['state'] . '"><b>' . $row['country'] . '</b></td>';
+      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: '; echo $row['city'] ?? 'n/a'; echo ' State: '; echo $row['state'] ?? 'n/a'; echo '"><b>' . $row['country'] . '</b></td>';
```
noplanman: This could also be made easier to read:
`title="' . sprintf('City: %1$s State: %2$s', $row['city'] ?? 'n/a', $row['state'] ?? 'n/a') . '"`

Or a mini simplification from what you have now, removing the extra `echo`s:
`title="City: ' . ($row['city'] ?? 'n/a') . ' State: ' . ($row['state'] ?? 'n/a') . '"`

dmorley: AH, I tried that, I missed the () to make it work so I did it the hard way

noplanman:
Merge train, coming through... 🚅

noplanman:
> hrm what is this new box, am I talking to myself now?

You talking about the "Review" box? A nice new feature to link multiple comments together to review a PR 👍

PR #66, opened 2017-01-12
Fixsql

dmorley:
Round 1


cleanup.php
```diff
@@ -81,18 +81,18 @@
       $classver = 'black';
     }
     echo '<td class="' . $classver . '"><div title="' . $pre . ' codename: ' . $row['shortversion'] . ' master version is: ' . $row['masterversion'] . '" class="tipsy">' . $version . '</div></td>';
-    echo '<td>' . $row['uptimelast7'] . '</td>';
-    echo '<td>' . $row['responsetimelast7'] . '</td>';
-    echo '<td>' . ($row['signup'] === '1' ? 'Open' : 'Closed') . '</td>';
+    echo '<td>' . $row['uptime_alltime'] . '</td>';
+    echo '<td>' . $row['responsetimems'] . '</td>';
+    echo '<td>';  echo $row['signup'] ? 'Open' : 'Closed'; echo '</td>';
```
dmorley: I think this should be echo '<td>' . ($row['signup'] ? 'Open' : 'Closed') . '</td>';
Now that you know how to do this..

show.php
```diff
@@ -8,7 +8,7 @@
 $dbh = pg_connect("dbname=$pgdb user=$pguser password=$pgpass");
 $dbh || die('Error in connection: ' . pg_last_error());
 
-$sql = "SELECT * FROM pods WHERE adminrating <> -1 AND hidden <> 'yes' AND signup = 1 ORDER BY uptimelast7 DESC";
+$sql = "SELECT * FROM pods WHERE adminrating <> -1 AND NOT hidden AND signup ORDER BY uptime_alltime DESC";
```
noplanman: Not 100% sure (maybe 99%), but like in other languages, `if` conditions are evaluated left to right. So as soon as a condition fails, it moves to the next one.
For that reason, it (almost) always makes sense to move the lighter comparisons/operations to the front of the query/where/if statement.

Comparing if something is true or false is quicker than evaluating an integer value.

In this particular case it probably makes close to no difference, but I thought I'd take the opportunity to explain this here.

More info about this [here](https://en.wikipedia.org/wiki/Short-circuit_evaluation) and [here](https://secure.php.net/manual/en/language.expressions.php#21750).

show.php
```diff
@@ -32,14 +32,12 @@
 
   <?php
   while ($row = pg_fetch_array($result)) {
-    if ($row['secure'] === 'true') {
```
noplanman: Have you tested these changes?

Boolean values return `t` or `f` strings (which is super weird in my view, but hey...)
[Look!](https://www.postgresql.org/docs/9.2/static/datatype-boolean.html#DATATYPE-BOOLEAN-EXAMPLE)

Evaluating a string always returns `true` (unless it's an empty string)

dmorley: Hrm, I am using them locally with my DB with 2 test pods, let me dump from prod and see how things look. Is the way it was the best way to evaluate bool?

db/tables.sql
```diff
@@ -3,49 +3,46 @@ CREATE TABLE pods (
...
  total_users int, 
  active_users_halfyear int,
  active_users_monthly int,
  local_posts int,
- uptimelast7 numeric(5,2),
+ uptime_alltime numeric(5,2),
+ uptime_custom text,
  status text,
- responsetimelast7 text,
+ responsetimems text,
```
dmorley: Yea, I changed it, much clearer to everyone else without the ms

db/pull.php
```diff
@@ -291,13 +291,13 @@
         $uptime      = isset($matchper[1][0]) ? preg_replace('/,/', '.', $matchper[1][0]) : 0;
         $statslastdate = date('Y-m-d H:i:s');
         if (strpos($pingdom, "class=\"up\"")) {
-          $live = 'up';
+          $status = 'up';
```
noplanman: Nice change of variable name, much more descriptive! :tada:

db/tables.sql
```diff
@@ -3,49 +3,46 @@ CREATE TABLE pods (
...
  softwarename text,
+ masterversion text,
+ shortversion text,
  score int DEFAULT 10,
  weightedscore numeric(5,2) DEFAULT 10,
  adminrating decimal DEFAULT 10,
  userrating decimal DEFAULT 10,
- whois text,
- hidden text DEFAULT 'yes',
+ hidden boolean DEFAULT 'yes',
```
dmorley: The key words TRUE and FALSE are the preferred (SQL-compliant) usage.  see it now, I missed that part and read all are equal

db/tables.sql
```diff
@@ -30,15 +30,15 @@ CREATE TABLE pods (
...
+ responsetime text,
  service_facebook boolean,
  service_twitter boolean,
  service_tumblr boolean,
  service_wordpress boolean,
  service_xmpp boolean,
  token text,
  tokenexpire timestamp,
- terms text,
+ terms text DEFAULT '/terms',
```
noplanman: Do all pods have terms? I guess they should, but I was just wondering if a pod doesn't have a link to the terms, this field "should" stay empty, otherwise this creates a broken link.

Maybe we set this value as the default input when podmins add their pod?

(Reminder to keep the DB as lean as possible. For this project it might not make a big impact, but with thousands of rows it can get bulky real quick.)

show.php
dmorley: Hrm, kind of makes sense, I guess I assumed the DB would act smart and decide the fastest way to do it :)

I removed that first evaluation now anyways, since adminrating is unused. (yes I should remove from table I think)

go.php
```diff
@@ -17,7 +17,10 @@
   $row = pg_fetch_all($result);
   $row || die('unknown url');
 
-  //Add click counter +1 for $row[0]['domain'] clicks in future, separate click table
+  $sql    = 'INSERT INTO clicks (domain, manualclick) VALUES ($1, $2)';
```
noplanman: Whenever somebody clicks the pod link, `manualclick` gets reset.
This should be an `UPDATE` statement, as the row should be added as soon as the pod gets added.

(same thing a few lines down)

show.php
dmorley: Also my next branch was going to be to rip all the http out of the code base. Might not be a need to make it perfect on this branch.

db/tables.sql
dmorley: Ah, that makes sense, let me just put it on the form as the default when you add a pod. 
I was thinking about existing pods, but for that I should do a manual injection not a default

show.php
noplanman: The problem here is that the way PostgreSQL returns boolean values, is as strings.
So every comparison should be comparing with `=== 't'` or `=== 'f'`.

Would be great if it would just return a bool, but it doesn't. Maybe it can be set with flags somehow, will need to look this up.

Basically, wherever a bool from the db is being used in an if conditional, it needs to be explicitly compared 😕

go.php
dmorley: I was planning on a table of clicks, like many rows and add them up where needed. I guess it would make more sense to pull the clicks and add to them so each pod has a single row not a ton. I guess that is why I created the table this way but then implemented differently

showmap.php
```diff
@@ -8,6 +8,9 @@
   if ($cords[0] === $country_code) {
     $lat  = $cords[1];
     $long = $cords[2];
+  } else {
+    $lat = 31;
+    $long = -99;
```
noplanman: Brilliant, this was waiting somewhere in my stash of code snippets too 👍

go.php
noplanman: Aaah, I see what you mean, thanks for elaborating.

Is there any special reason you'd want separate rows for this?

Otherwise it could also just be an extra field in the `pods` table, that just gets incremented each time. (Same applying to `autoclick`)

db/tables.sql
noplanman: > I was thinking about existing pods, but for that I should do a manual injection not a default

Exactly! You can even make a `curl` request before updating to make sure it's really there 😉

show.php
dmorley: I see hacks but I would rather make it right in php than add functions to DB http://stackoverflow.com/a/4721668

I will go back to t or f for now

go.php
dmorley: I guess my brain was thinking a separate table with a timestamp for each click would allow for some interesting long term data like clicks per month, week, year, etc. Kind of the same thing I was thinking with the stats table, some level of being able to graph something at some point vs just an overall number. I think initially just a number by summing up the table, longer term add some cooler statistics.

show.php
noplanman: Yes, it's just so weird to return a string when it should be a boolean. Well, they must have thought it through when implementing it...

:+1: for sticking to `=== 't'` and `=== 'f'` for now.

go.php
noplanman: Riiiight, ok! Then keep it that way!

showmap.php
dmorley: Only because https://localhost shows no map! I had to fix it.

noplanman: What I did in my local version, was to assign a default country code instead of an empty string to `$country_code`.

noplanman: Also, just FYI, the map doesn't show up because the Javascript generated [here](https://github.com/diasporg/Poduptime/blob/ffa4e65d32062131ee928bf67ab99e61eb046a9d/showmap.php#L70) is invalid if those values aren't set.

db/migration00001.sql
```diff
@@ -0,0 +1,36 @@
...
+ALTER TABLE pods ALTER COLUMN secure TYPE boolean USING secure::boolean;
+ALTER TABLE pods ALTER COLUMN signup TYPE boolean USING signup::boolean;
+
+DROP TABLE users;
+
+CREATE TABLE apikeys (
+ key text,
+ email text,
+ usage int,
+ dateCreated timestamp DEFAULT current_timestamp
```
noplanman: Just noticed the camelCase here and all date fields. Is this some PostgreSQL convention or something?
Maybe make this snake_case like the other columns?

This is mainly to not mix different styles.

db/pull.php
```diff
@@ -168,10 +168,20 @@
     if (!$output && !$outpulssl && !$domain) {
       continue;
       echo 'no connection to pod';
+      
+      $sql    = 'INSERT INTO checks (domain, online, error) VALUES ($1, $2)';
```
noplanman: ❗️ Missing the `$3` here

podmin.php
```diff
@@ -3,9 +3,10 @@
 Its easy start monitoring on your pod with a free <a href="https://uptimerobot.com" target="_blank">Uptime Robot</a> account.<br>
 Create a monitor for your pod, then in "My Settings" create a monitor-specific API key and paste below.<br>
 <br>
-<form action="https://podupti.me/db/add.php" method="post">
+<form action="https://<?php echo $_SERVER['HTTP_HOST'] ?>/db/add.php" method="post">
   <label>Monitor API Key: <input type="text" name="url" class="xlarge span8" placeholder="m58978-80abdb799f6ccf15e3e4ttwe"> (don't copy the period on the end)</label><br>
   <label>Pod Domain Name: <input type="text" name="domain" class="xlarge span4" placeholder="domain.com"></label><br>
+  <label>Pod Terms Link: <input type="text" name="domain" class="xlarge span4" value="/terms" placeholder="/terms"></label><br>
```
noplanman: ❗️ Ts ts ts, copy pastin' huh? 😋

db/migration00001.sql
dmorley: Blame copy and paste from examples I am sure, no actual reason.

podmin.php
dmorley: I WOULD NEVER

noplanman:
Really great PR, thanks! :tada:

dmorley:
Yea. Noticed the n/a on city and state not working properly https://dev.podupti.me/?advancedview did I get the php7 thing wrong?

noplanman:
Oh, just realised that I got the `??` a bit wrong!
I thought is was identical to `?:`, with the difference of not throwing a warning if the first value doesn't exist (like `$arr['val']` that hasn't been defined).

BUT, it only does exactly that and checks for `null`, not `empty()` (which `?:` does).

In this case, you can change it to `?:`, because the value will be defined, coming from the db.

dmorley:
Been trying to read and debug also, but something still does not make sense:
php:
$city    = isset($location['city']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['city']) : null;

postgres:
podupdb=> SELECT city FROM pods WHERE city IS NOT NULL;
  city  
--------
 
 Latham
(2 rows)

podupdb=> SELECT city FROM pods WHERE city IS NULL;
 city 
------
(0 rows)


So even the city with no results is not setting as NULL in the DB, but postgresql represents NULL as blank supposedly.

If the DB was returning the values as NULL then the ?? would work properly in PHP, so I think I am not actually setting null when I think I am.

dmorley:
Ah, It is my isset
Should be 
$city    = !empty($location['city']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['city']) : null;

The output of the data is:
..
  ["city"]=>
  string(0) ""
..

so it is "set" but it is also empty

noplanman:
You obviously have more than 2 rows in the db, right?

It works properly for the version of the db that I have now, where all cities are null.
Yes, I think the PHP implementation returns an empty string instead of null, because otherwise `??` would work, as you say.

noplanman:
Oooh, missed that one, nice catch!! :+1:

PR #69, opened 2017-01-13
Fixes

dmorley:


db/pull.php
```diff
@@ -341,8 +341,7 @@
       $uptimerobotstat = $uptr->stat;
       $uptime          = $uptr->monitors->monitor{'0'}->alltimeuptimeratio;
       $uptime_custom   = $uptr->monitors->monitor{'0'}->customuptimeratio;
-      $diff            = abs(strtotime(date('Y-m-d H:i:s')) - strtotime($dateadded));
-      $months          = floor(($diff - $years * 365 * 60 * 60 * 24) / (30 * 60 * 60 * 24));
+      $months          = ((new DateTime())->diff(new DateTime($dateadded))->m + ((new DateTime())->diff(new DateTime($dateadded))->y*12));
```
noplanman: To minimise on code and calculation (which will totally save half a pico-second), this is nicer:
```php
$diff            = (new DateTime())->diff(new DateTime($dateadded));
$months          = $diff->m + ($diff->y * 12);
```

show.php
```diff
@@ -54,7 +54,7 @@
     echo '<td>' . $row['uptime_alltime'] . '%</td>';
     echo '<td data-toggle="tooltip" data-placement="bottom" title="active six months: ' . $row['active_users_halfyear'] . ', active one month: ' . $row['active_users_monthly'] . '">' . $row['active_users_halfyear'] . '</td>';
     if ($country_code === $row['country']) {
-      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: '. ($row['city'] ?? 'n/a') . ' State: ' . ($row['state'] ?? 'n/a') . '"><b>' . $row['country'] . '</b></td>';
+      echo '<td class="text-success" data-toggle="tooltip" data-placement="bottom" title="City: ' . ($row['city'] ?? 'n/a') . ' State: ' . ($row['state'] ?? 'n/a') . '"><b>' . $row['country'] . '</b></td>';
     } else {
       echo '<td data-toggle="tooltip" data-placement="bottom" title="City: '. ($row['city'] ?? 'n/a') . ' State: ' . ($row['state'] ?? 'n/a') . '">' . $row['country'] . '</td>';
```
noplanman: Might as well fix this one too :blush:

rate.php
```diff
@@ -69,7 +69,7 @@
     if ($row['admin'] === '1') {
```
noplanman: Is this supposed to be a `boolean` field too?

dmorley:
The whole admin/adminrating/admincomments thing never got used and I am not sure its a feature worth keeping. I should have gutted it with the sqlchanges, but I was still pondering if it had any use. If it stays then yes this should be bool. Thoughts on if it's a feature worth keeping around?

January 13, 2017 9:32 AM, "Armando Lüscher"  wrote:
	@noplanman commented on this pull request. 
------------------------------------
	In rate.php (https://github.com/diasporg/Poduptime/pull/69#pullrequestreview-16578898): 

	> @@ -69,7 +69,7 @@ if ($row['admin'] === '1') { 

	Is this supposed to be a boolean field too?

noplanman:
Could you explain what it was/is for exactly?

Are these comments made by user for/about the admin, or actual admin comments? If so, what for?

dmorley:
Comments and ratings made by admin. I do not remember why, I just did it when I added user comments. I am way to lazy to comment on pods, and I think since the system does more and more checking to see how valid a pod is, maybe I felt like a "bad" pod would be listed somehow. The whole comment thing should really be gutted, no one leaves them anymore, if they do its usually the podmin. Might make more sense to just have podmins able to say something about their pod to the public.

PR #74, opened 2017-01-15
Populate sslexpire, dnssec..

dmorley:
Not sure how to show more data on pods, the table is already big, the text already big, more data great but how to show, no big ideas from me.

Also not sure if there is an easier way to parse the delv info, I tried a lot of things, but since the needed data is in the first ; line it gets filtered out easily it seems


dmorley:
&& git rebase -i HEAD~4 worked for me here, but when I try and push it says already current.

noplanman:
Might be an idea to start working with images to shorten the table.

- SSL could be a little lock next to the domain name
- Terms could be a little scroll or something
- Uptime cut the decimals, i.e. (100.00% -> 100%)
- Score put the `/100` into the header, remove from all cells in the column
- Rating put the `/10` into the header, remove from all cells in the column
- Boolean values (like IPv6 and Signups) could be a tick or cross image.

Just a few ideas to shrink the table. But I guess if somebody is looking at the advanced table, advanced table is what they shall get! :wink:

About the rebase command. An easier option is to use `git rebase -i develop`, which lists all changes made since develop.
When you squash commits that have already been pushed, you need to force the push to overwrite the upstream commits, like so: `git push -f`

db/pull.php
```diff
@@ -187,7 +191,10 @@
       echo '<br>Signup Open: ' . $signup . '<br>';
     }
     $ip6    = exec(escapeshellcmd('dig +nocmd ' . $domain . ' aaaa +noall +short'));
-    $ip     = exec(escapeshellcmd('dig +nocmd ' . $domain . ' a +noall +short'));
+    exec(escapeshellcmd('delv ' . $domain), $iplookup);
+    $dnssec = str_replace("; ","",$iplookup[0]) === 'fully validated' ? t : f ;
+    preg_match('/A(.*)/', $iplookup[1], $version);
+    $ip = trim($version[1]);
     $test   = strpos($ip6, ':');
     $ipv6   = $test === false ? 'no' : 'yes';
```
noplanman: Might as well merge `$test` away, as it's only being used here:
```php
$ipv6   = strpos($ip6, ':') === false ? 'no' : 'yes';
```

db/pull.php
```diff
@@ -187,7 +191,10 @@
       echo '<br>Signup Open: ' . $signup . '<br>';
     }
     $ip6    = exec(escapeshellcmd('dig +nocmd ' . $domain . ' aaaa +noall +short'));
-    $ip     = exec(escapeshellcmd('dig +nocmd ' . $domain . ' a +noall +short'));
+    exec(escapeshellcmd('delv ' . $domain), $iplookup);
+    $dnssec = str_replace("; ","",$iplookup[0]) === 'fully validated' ? t : f ;
```
noplanman: Would it work to just look for the substring `fully validated` here?

Also, `delv` doesn't seem to be a common command, right?
Can't find it on my Mac (or on brew), nor the debian repos.
Where do I get it? Tried installing `bind9`, but no luck.

dmorley: Yea that would be easier!

delv replaced dig in 2014 for dnssec usage, dig does have +dnssec but that will not validate that the sigs are correct just that they exist. 

https://kb.isc.org/article/AA-01152/0/Eleven-twelve-dig-and-delv%3A-BIND-9.10.html

Ubuntu has had it for years, I assumed debian also then since they just copy and paste most things :call_me_hand: 

What version of bind9 is your default?

noplanman: My version on Ubuntu is 9.8.1, just saw that the post you linked is talking about 9.10. What source do I need to apt into to get 9.10?

Just saw that it needed a `brew install bind` for Mac, not `bind9`.

dmorley:
I was also thinking that "services" could be more than external services. Like ipv6 and dnssec could be part of that icon set and be marked as services but I find that harder to use if you are looking for something. Maybe you are right and the table can just be big! I like data and the idea of the full table is to be able to make a full decision on a pod. If you want simple we have that as the default.

noplanman:
What would make sense though, it to be able to filter it, check this example:
https://thatoneprivacysite.net/vpn-comparison-chart/

db/pull.php
dmorley: bind9 has been 9.10 since xenial, so it is newer than I thought, odd since 9.10 is EOL already as 9.11 is stable.  That means your on precise for ubuntu?

noplanman: Ah yes, the vagrant box I have is on precise. For testing I'll just comment that block of code, no problem.

noplanman:
@dmorley Do you want to keep this PR open for the whole table redesign or will you create a separate PR for that?

dmorley:
Let me do basic cleanup here, then a new for a better cooler table. What do you think about some checks and stuff for now?

noplanman:
What do you mean by `checks and stuff`?

dmorley:
literally utf8 checkmarks and links symbols.

Did I mess up the rebase here? It shows all your work as part of this PR now?

noplanman:
Something in the rebase got messed up, yes.
Should be able to merge though, as all changes are already in develop anyway.

You want to add anything or can we merge?

dmorley:
Go for it now if looks ok. I clearly need to learn how to rebase, it just took me 5 --continues and it's a disaster

noplanman:
Have cleaned this branch up, how in the world did you do that?! haaha 😄 
Need to find a way to prevent this in the future.

Will let you merge.

PR #79, opened 2017-01-16
Fixpullnbackup

dmorley:
I wonder how this will test for you. 

Also I realized that this could be a cron.php and I think be able to do all these things just fine?


db/pull.sh
```diff
@@ -16,8 +16,8 @@ if [ ! -s /tmp/index.google ];then
 exit;
 else
 	echo "Pulling in new pod data";
-	cd /var/www/poduptime/db
-	php pull.php debug=1
+	cd $SCRIPT_DIR
```
noplanman: Add quotes here, in case there are whitespaces in the path.

dmorley:
I do, I do not know how to solve the warnings tbh, fix away!

db/pull.sh
dmorley: :ok:

noplanman:
As I'm fixing the warning and errors, I've opened pandora's box ❗️

The issue I found is [this if statement](https://github.com/diasporg/Poduptime/blob/454e55560f583a6e7a12fe2366b2dc9e6145e5e5/db/pull.php#L150).

What should be done if a pod can't be connected to?

This is a pretty difficult question, as it can be due to a whole bunch of factors.

I think a good move would be to have a counter that remembers how many failed attempts there were and a when limit is reached, the poduptime admin gets notified about it.

As for the pod itself, I think it would suffice to set the response time / ping to "n/a" or something, to let the user know that it wasn't accessible during the last pull.

What's your take on this?

dmorley:
I think my thought with that if statement was to validate the json, but now with nodeinfo its not just a random location with json data, so that is super old code, your changes do the same thing.

In the past when a pod could not be connected to it would really mess up results, I made some attempts to fix that but its not right. Really if your a perfect pod and one time it can not connect it should just leave your stats in pod table be, and then mark you in the checks table as missing a check, no big deal to miss one or two. So I think the counter you are thinking of is the checks table already?

PR #81, opened 2017-01-16
Fixpullnbackup fixes

noplanman:
This PR attempts to fix most warnings and errors and optimises some parts of the code.

@dmorley I know this is a big PR, but most of the changes are visual. Best to look at each individual commit and skip the one marked as visual only.


noplanman:
I've added a little `_debug` function to clean up all the debugging code and make it 1 line per debug statement.

dmorley:
Nice as you can see in first lines I struggled a bit with debug. I was trying to make it so you can call debug via cli and fpm but I never won that fight. I recently found that you can set GET values from cli with php-cgi

noplanman:
Right, I'll add a debug parameter then, as it's just set to true at the moment...

noplanman:
Ok, debug parameter added :+1:

dmorley:
Looks good, How does this work with my PR? Do that then this?

noplanman:
First merge mine into yours, then yours into develop.

db/pull.php
```diff
@@ -37,262 +35,237 @@
...
+  if ($location) {
+    $country = !empty($location['country_code']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['country_code']) : null;
+    $city    = !empty($location['city']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['city']) : null;
+    $state   = !empty($location['region']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['region']) : null;
+    $lat     = !empty($location['latitude']) ? $location['latitude'] : null;
+    $long    = !empty($location['longitude']) ? $location['longitude'] : null;
+
+    // If lat and long are just a generic country with no detail lets make some tail up
+    // else openmap just stacks them all on top of each other.
+    strlen($lat) < 4 && $lat += (random_int(1, 15) / 10);
```
noplanman: @dmorley Is this really still necessary?
It seems mapbox does a really pretty job.
mapbox-markersmapbox-markers-spiral

dmorley:
For sure, that hack was a long time ago

dmorley:
As in for sure its no longer needed

noplanman:
Ok, have pushed their removal.

noplanman:
Just because one should always make the most out of a bug :tada:
marker-awesomeness

dmorley:
Ha, frame it!

db/pull.php
```diff
@@ -37,262 +35,232 @@
...
-    }
-    if ($location) {
-      $country = !empty($location['country_code']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['country_code']) : null;
-      $city    = !empty($location['city']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['city']) : null;
-      $state   = !empty($location['region']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['region']) : null;
-      $lat     = !empty($location['latitude']) ? $location['latitude'] : null;
-      $long    = !empty($location['longitude']) ? $location['longitude'] : null;
+  if (!$jsonssl && !$domain) {
+    continue;
+    /*echo 'no connection to pod';
```
dmorley: I just noticed this commented on on my pull.sh run, no good?

noplanman: Well, it has the `continue` just above it, so this never gets executed!

I commented it out to make it easier to read, to make sure it wasn't mistaken for being executed 😄 

The `continue` should probably come after the SQL stuff. Then it can be uncommented :+1:

PR #84, opened 2017-01-17
fix some pull issues

dmorley:
Getting lots of inconsistancy with dnssec, so I hardcoded the IP pull for testing it out for a while. And score heavy down if your dnssec is setup and invalid, since that is the idea of dnssec.

Google DNS does not even respond and give a record if dnssec invalid
Open DNS does not send back RRSIG's so all records are invalid.

I can not seem to set the CURL options with DNS servers, that is not a thing. Of course I use different DNS servers all over the place, and my dev environment has different ones constantly.

### DIFF
--- a/db/pull.php
+++ b/db/pull.php
@@ -170,13 +170,8 @@ while ($row = pg_fetch_assoc($result)) {
   $statslastdate = date('Y-m-d H:i:s');
   $ping          = curl_init();
   curl_setopt($ping, CURLOPT_URL, 'https://api.uptimerobot.com/getMonitors?format=json&noJsonCallback=1&customUptimeRatio=7-30-60-90&responseTimes=1&responseTimesAverage=86400&apiKey=' . $row['stats_apikey']);
-  curl_setopt($ping, CURLOPT_POST, 0);
-  curl_setopt($ping, CURLOPT_HEADER, 0);
   curl_setopt($ping, CURLOPT_RETURNTRANSFER, 1);
   curl_setopt($ping, CURLOPT_CONNECTTIMEOUT, 8);
-  curl_setopt($ping, CURLOPT_NOBODY, 0);
-  curl_setopt($ping, CURLOPT_MAXCONNECTS, 5);
-  curl_setopt($ping, CURLOPT_FOLLOWLOCATION, true);
   $uptr = json_decode(curl_exec($ping));
   curl_close($ping);
   _debug('Uptime Robot', $uptr, true);
--- a/db/pull.php
+++ b/db/pull.php
@@ -23,7 +23,7 @@ preg_match('/number: "(.*?)"/', $outputmv, $version);
 $dmasterversion = trim($version[1], '"');
 _debug('Diaspora Masterversion', $dmasterversion);
 
-//get master code version for freindica pods
+//get master code version for friendica pods
 $mv = curl_init();
 curl_setopt($mv, CURLOPT_URL, 'https://raw.githubusercontent.com/friendica/friendica/master/boot.php');
 curl_setopt($mv, CURLOPT_POST, 0);
@@ -43,11 +43,9 @@ $dbh || die('Error in connection: ' . pg_last_error());
 //foreach pod check it and update db
 if ($_domain) {
   $sql = 'SELECT domain,stats_apikey,score,date_created,adminrating,weight FROM pods WHERE domain = $1';
-  //$sleep  = '0';
   $result = pg_query_params($dbh, $sql, [$_domain]);
 } elseif (PHP_SAPI === 'cli') {
   $sql = 'SELECT domain,stats_apikey,score,date_created,adminrating,weight FROM pods';
-  //$sleep  = '1';
   $result = pg_query($dbh, $sql);
 } else {
   die('No valid input');
@@ -91,11 +89,11 @@ while ($row = pg_fetch_assoc($result)) {
   curl_setopt($chss, CURLOPT_RETURNTRANSFER, 1);
   curl_setopt($chss, CURLOPT_NOBODY, 0);
   curl_setopt($chss, CURLOPT_CERTINFO, 1);
-  curl_setopt($chss, CURLOPT_VERBOSE, 1);
+  curl_setopt($chss, CURLOPT_VERBOSE, 0);
   $outputssl      = curl_exec($chss);
   $outputsslerror = curl_error($chss);
   $info           = curl_getinfo($chss, CURLINFO_CERTINFO);
-  $sslexpire      = $info[0]['Expire date'];
+  $sslexpire      = $info[0]['Expire date'] ?? null;
   curl_close($chss);
 
   _debug('Nodeinfo output', $outputssl, true);
@@ -104,13 +102,13 @@ while ($row = pg_fetch_assoc($result)) {
   //get new json from nodeinfo
   $jsonssl = json_decode($outputssl);
 
-  if (!$jsonssl && !$domain) {
-    continue;
-    /*echo 'no connection to pod';
+  if (!$jsonssl && !$domain) {    
+    _debug('Connection:', 'Can not connect to pod');
 
     $sql    = 'INSERT INTO checks (domain, online, error) VALUES ($1, $2, $3)';
     $result = pg_query_params($dbh, $sql, [$domain, false, $outputsslerror]);
-    $result || die('Error in SQL query: ' . pg_last_error());*/
+    $result || die('Error in SQL query: ' . pg_last_error());
+    continue;
   }
 
   $sql_checks    = 'INSERT INTO checks (domain, online) VALUES ($1, $2)';
@@ -121,7 +119,7 @@ while ($row = pg_fetch_assoc($result)) {
   if ($jsonssl !== null) {
     $score += 1;
 
-    var_dump($jsonssl);
+    _debug('Json raw data', $jsonssl, true);
     if ($jsonssl->openRegistrations === true) {
       $signup = true;
     }
@@ -142,7 +140,7 @@ while ($row = pg_fetch_assoc($result)) {
     $service_twitter       = in_array('twitter', $jsonssl->services->outbound, true);
     $service_tumblr        = in_array('tumblr', $jsonssl->services->outbound, true);
     $service_wordpress     = in_array('wordpress', $jsonssl->services->outbound, true);
-    $service_xmpp          = $jsonssl->metadata->xmppChat === true;
+    $service_xmpp          = $jsonssl->metadata->xmppChat === true ?? 'f';
   } else {
     $score -= 1;
     $dver         = '.connect error';
@@ -151,14 +149,18 @@ while ($row = pg_fetch_assoc($result)) {
   }
 
   _debug('Signup Open', $signup);
-  $ip6 = exec(escapeshellcmd('dig +nocmd ' . $domain . ' aaaa +noall +short'));
+  $ip6 = exec(escapeshellcmd('dig @4.2.2.2 +nocmd ' . $domain . ' aaaa +noall +short'));
   $iplookup = [];
-  exec(escapeshellcmd('delv ' . $domain), $iplookup);
+  exec(escapeshellcmd('delv @4.2.2.2 ' . $domain), $iplookup);
   $dnssec = in_array('; fully validated', $iplookup);
-  preg_match('/A(.*)/', $iplookup[1], $version);
-  $ip   = trim($version[1]);
+  if ($iplookup) {
+    preg_match('/A(.*)/', $iplookup[1], $version);
+    $ip   = trim($version[1]);
+  }
+  $ip || $score -= 4;
   $ipv6 = strpos($ip6, ':') !== false;
   _debug('IP', $ip);
+  _debug('IPv6', $ip6);
 
   $location = geoip_record_by_name($ip);
   _debug('Location', $location, true);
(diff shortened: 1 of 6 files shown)


noplanman:
Maybe we can make use of this library?
https://packagist.org/packages/metaregistrar/php-dnssec-validator

db/pull.php
```diff
@@ -104,13 +102,13 @@
   //get new json from nodeinfo
   $jsonssl = json_decode($outputssl);
 
-  if (!$jsonssl && !$domain) {
-    continue;
-    /*echo 'no connection to pod';
+  if (!$jsonssl && !$domain) {
```
noplanman: Probably best to move the `$outputsslerror` debug message further up into here, as this is only necessary if something goes wrong.

Also, `$domain` is always set here, right? As it's coming straight from the DB.
Unless the DB has an entry with no domain set, which shouldn't be possible, right?

db/pull.php
```diff
@@ -121,7 +119,7 @@
   if ($jsonssl !== null) {
     $score += 1;
 
-    var_dump($jsonssl);
+    _debug('Json raw data', $jsonssl, true);
```
noplanman: I left this cause I wasn't sure if this was part of debug output or not.
This is basically exactly the same as `$outputssl` above, just JSON encoded. Maybe we can remove one of the outputs? Or if not (which could make sense to keep both), at lease output them right after each other, to keep them together in context. Reason for keeping both would be if the JSON string has an error in it and `$jsonssl` is `null` because of that.

Also, all curl requests should also only output something when in debug mode, right? At the moment they output all the things I think.

db/pull.php
```diff
@@ -151,14 +149,18 @@
...
   _debug('Signup Open', $signup);
-  $ip6 = exec(escapeshellcmd('dig +nocmd ' . $domain . ' aaaa +noall +short'));
+  $ip6 = exec(escapeshellcmd('dig @4.2.2.2 +nocmd ' . $domain . ' aaaa +noall +short'));
   $iplookup = [];
-  exec(escapeshellcmd('delv ' . $domain), $iplookup);
+  exec(escapeshellcmd('delv @4.2.2.2 ' . $domain), $iplookup);
   $dnssec = in_array('; fully validated', $iplookup);
-  preg_match('/A(.*)/', $iplookup[1], $version);
-  $ip   = trim($version[1]);
+  if ($iplookup) {
```
noplanman: Since `$dnssec` also uses `$iplookup`, might as well take that assignment into here too.

Also, remember to set default values for `$dnssec` and `$ip` before the if condition!

db/pull.php
```diff
@@ -142,7 +140,7 @@
     $service_twitter       = in_array('twitter', $jsonssl->services->outbound, true);
     $service_tumblr        = in_array('tumblr', $jsonssl->services->outbound, true);
     $service_wordpress     = in_array('wordpress', $jsonssl->services->outbound, true);
-    $service_xmpp          = $jsonssl->metadata->xmppChat === true;
+    $service_xmpp          = $jsonssl->metadata->xmppChat === true ?? 'f';
```
noplanman: Stick to booleans here! Instead of `'f'` just use `false`.

The boolean is correctly cast when inserting into the DB.

db/pull.php
```diff
@@ -184,33 +186,32 @@
...
+  if ($uptr->stat === 'ok') {
+    $responsetime    = $uptr->monitors->monitor{'0'}->responsetime{'0'}->value ?? 'n/a';
+    $uptimerobotstat = $uptr->stat;
+    $uptime          = $uptr->monitors->monitor{'0'}->alltimeuptimeratio;
+    $uptime_custom   = $uptr->monitors->monitor{'0'}->customuptimeratio;
+    $diff            = (new DateTime())->diff(new DateTime($dateadded));
+    $months          = $diff->m + ($diff->y * 12);
+    if ($uptr->monitors->monitor{'0'}->status == 2) {
+      $status = 'Up';
+      $score += 2;
```
noplanman: Just FYI.

Before: an error reduced the score by 2.
Now: success increases the score by 2.

Is this change deliberate?

db/pull.php
dmorley: Right, I like the json but I would rather have the error output also, it's very handy to debug.

I set the curl verbose to off, I had only enabled as I read it must be enabled for sslcert data, that is untrue at test.

db/pull.php
dmorley: I think so, it was testing funny on pulls, and now it is testing properly. Some things were going up that actually were failing on manual check.

db/pull.php
dmorley: Makes sense for $dnssec, but for $ip I am trying to solve for delv queries that do not return an A record, line 159, as they were getting scored as good without since they would resolve for curl but not for delv.

dmorley:
I gave that package a run before switching to delv:

    Clone this repository
    php test.php domainname

Ends in error, and searching the error did not yeild any results, so I moved on, I did not want to use something that looks kind of abandonded and did not work out of the box

I also tried https://netdns2.com/ package, it worked, but was of course setup with pear and had some setup issues I had to resolve.

So I opted for delv since everyone has it! :rofl:

db/pull.php
dmorley: This is not working properly since sometimes this value is just simply missing, I mean it works but then a PHP Warning, would be nice to get rid of them. I could do it with a $if ($jsonssl->metadata->xmppChat) {} but that seems flawed, so thinking on it

db/pull.php
dmorley: ah I removed $domain and it broke, I see now the error is:
Query failed: ERROR:  invalid input syntax for type boolean: "" in /db/pull.php on line 109
This makes zero sense, the boolean value is set to false right there

dmorley: and it does not in fact "continue" the loop dies on the first pod that is bad. I tried 'false' and now the pgerror is gone, but why the f does true work a few rows down without. Something crazy here.

db/pull.php
noplanman: > I would rather have the error output also

Yes of course, that's not what I'm talking about. I mean the output of the curl response, which is the JSON string. Further up, the string itself is output, and here the JSON encoded object, which in essence is exactly the same thing.
Unless the JSON string is malformed, in which case the object would be `null`.

Absolutely keep the error output!

db/pull.php
noplanman: Yes, I understand. What I meant was this:
```php
$service_xmpp          = $jsonssl->metadata->xmppChat === true ?? false;
```

To set `false` in case the value doesn't exist.

noplanman:
Getting Net_DNS2 from packagist would be super simple! No need to install anything special, as it's pure PHP.

Are you familiar with composer and package/dependency management?

dmorley:
Oh, I see my sample also for Net_DNS2, it does not do dnssec validation so it was useless to me. 

No I have never used composer.

noplanman:
Have you seen this SO post?
https://stackoverflow.com/questions/29147354/how-to-validate-dnssec-in-php

You'll get to know composer soon enough :wink:

dmorley:
Yup that is how I went down that path. It will give your RRSIGS and stuff, but you then need a way to go up the line and validate they are valid sigs, it does not look to have that feature set.

db/pull.php
noplanman: Boolean fields can't handle strings. If you directly pass a bool, it works, otherwise it first has to be cast to an integer (have a look at the very last SQL query at the very bottom of the file.

So here you could change `'false'` to `false`, which should do it. Has nothing to do with `$domain` though.

dmorley: Understood, however it's doing to opposite. Will push up now without the 'false' and let you take a look. Something crazy, good chance it's me. false gets error, 'false' works now.

noplanman:
Ok, I know too little about all this DNS stuff to help you out 😕

dmorley:
ok, I figured out the true vs false in php/postgres: https://bugs.php.net/bug.php?id=33876

What crazyness.

db/pull.php
```diff
@@ -106,29 +120,21 @@
...
-    }
+    
+    (!$jsonssl->software->version) || $score += 1;
     $xdver        = $jsonssl->software->version ?? 0;
     $dverr        = explode('-', trim($xdver));
     $shortversion = $dverr[0];
     _debug('Version code', $shortversion);
-    $shortversion || $score -= 2;
-
+    $signup                = ($jsonssl->openRegistrations === true) ? 't' : 'f';
```
noplanman: Regarding the whole PostgreSQL mess that PHP does, I think for us to stay sane, we should stick with 1 way of handling this.
Either we do boolean strings with 't' and 'f', or we just stick to booleans and then cast to integer when adding to the DB query (like the last one).

I personally prefer the boolean then cast option, as it makes life so much easier to just play with booleans during the script instead of `=== 't'` or `=== 'f'`.

noplanman:
Well done on finding the PostgreSQL issue 👍 
Sorry I missed it!

Have you ever looked into PDO? Maybe that works better instead of the native PHP/PostgreSQL stuff?

db/pull.php
```diff
@@ -37,17 +37,29 @@
...
+$mv = curl_init();
+curl_setopt($mv, CURLOPT_URL, 'https://raw.githubusercontent.com/redmatrix/hubzilla/master/boot.php');
+curl_setopt($mv, CURLOPT_POST, 0);
+curl_setopt($mv, CURLOPT_HEADER, 0);
+curl_setopt($mv, CURLOPT_CONNECTTIMEOUT, 5);
+curl_setopt($mv, CURLOPT_RETURNTRANSFER, 1);
+curl_setopt($mv, CURLOPT_NOBODY, 0);
+$outputmv = curl_exec($mv);
+curl_close($mv);
+preg_match('/STD_VERSION\',             \'(.*?)\'/', $outputmv, $version);
```
noplanman: Just noticed this. It's better to do something more flexible! If they add an extra space or reformat this line, the script fails.

Compare with this for example:
```php
// Hubzilla
preg_match('/define.*\'STD_VERSION\'.*\'(.*?)\'/', $outputmv, $version);

// Friendica
preg_match('/define.*\'FRIENDICA_VERSION\'.*\'(.*?)\'/', $outputmv, $version);

// diaspora*
preg_match('/number:.*"(.*?)"/', $outputmv, $version);
```

dmorley: Yea better!

dmorley:
Ok, done!

db/pull.php
```diff
@@ -107,34 +107,37 @@
...
   //get new json from nodeinfo
   $jsonssl = json_decode($outputssl);
 
   if (!$jsonssl) {    
     _debug('Connection', 'Can not connect to pod');
 
-    $sql_errors    = 'INSERT INTO checks (domain, online, error) VALUES ($1, $2, $3)';
-    $result_errors = pg_query_params($dbh, $sql_errors, [$domain, 'f', $outputsslerror]);
+    $sql_errors    = 'INSERT INTO checks (domain, online, error, ttl) VALUES ($1, $2, $3, $4)';
+    $result_errors = pg_query_params($dbh, $sql_errors, [$domain, (int) false, $outputsslerror, $ttl]);
```
noplanman: Last thing, you might as well just put `0` here instead of `(int) false`, as it's a fixed `false` anyway.

db/pull.php
```diff
@@ -107,34 +107,37 @@
...
     $result_errors || die('Error in SQL query: ' . pg_last_error());
   }
 
   if ($jsonssl !== null) {
-    $sql_checks    = 'INSERT INTO checks (domain, online) VALUES ($1, $2)';
-    $result_checks = pg_query_params($dbh, $sql_checks, [$domain, 't']);
+    $sql_checks    = 'INSERT INTO checks (domain, online, ttl) VALUES ($1, $2, $3)';
+    $result_checks = pg_query_params($dbh, $sql_checks, [$domain, (int) true, $ttl]);
```
noplanman: Same here, just use `1` instead of `(int) true`.

db/pull.php
```diff
@@ -107,34 +107,37 @@
...
     $result_checks || die('Error in SQL query: ' . pg_last_error());
     
     (!$jsonssl->software->version) || $score += 1;
     $xdver        = $jsonssl->software->version ?? 0;
     $dverr        = explode('-', trim($xdver));
     $shortversion = $dverr[0];
     _debug('Version code', $shortversion);
-    $signup                = ($jsonssl->openRegistrations === true) ? 't' : 'f';
+    $signup                = ($jsonssl->openRegistrations === true) ? true : false;
```
noplanman: The conditional is already a bool, so this can be simplified to just:
```php
$signup                = $jsonssl->openRegistrations === true;
```

db/pull.php
```diff
@@ -237,7 +242,7 @@
...
-  $result_set  = pg_query_params($dbh, $sql_set, [$domain, 1, (int) $hidden, $ip, (int) $ipv6, $months, $uptime, $status, $statslastdate, $timenow, $responsetime, $score, $admin_rating, $country, $city, $state, $lat, $long, $user_rating, $shortversion, $masterversion, $signup, $total_users, $active_users_halfyear, $active_users_monthly, $local_posts, $name, $comment_counts, (int) $service_facebook, (int) $service_tumblr, (int) $service_twitter, (int) $service_wordpress, $weightedscore, (int) $service_xmpp, $softwarename, $outputsslerror, $uptime_custom, (int) $dnssec, $sslexpire]);
+  $result_set  = pg_query_params($dbh, $sql_set, [$domain, 1, (int) $hidden, $ip, (int) $ipv6, $months, $uptime, $status, $statslastdate, $timenow, $responsetime, $score, $admin_rating, $country, $city, $state, $lat, $long, $user_rating, $shortversion, $masterversion, (int) $signup, $total_users, $active_users_halfyear, $active_users_monthly, $local_posts, $name, $comment_counts, (int) $service_facebook, (int) $service_tumblr, (int) $service_twitter, (int) $service_wordpress, $weightedscore, (int) $service_xmpp, $softwarename, $outputsslerror, $uptime_custom, (int) $dnssec, $sslexpire]);
```
noplanman: Ooooh, nice catch!!

db/pull.php
```diff
@@ -12,42 +12,45 @@
...
-curl_setopt($mv, CURLOPT_POST, 0);
-curl_setopt($mv, CURLOPT_HEADER, 0);
 curl_setopt($mv, CURLOPT_CONNECTTIMEOUT, 5);
 curl_setopt($mv, CURLOPT_RETURNTRANSFER, 1);
-curl_setopt($mv, CURLOPT_NOBODY, 0);
 $outputmv = curl_exec($mv);
 curl_close($mv);
-preg_match('/FRIENDICA_VERSION\',      \'(.*?)\'/', $outputmv, $version);
+preg_match('/define.*\'FRIENDICA_VERSION\'.*\'(.*)\'/', $outputmv, $version);
 $fmasterversion = trim($version[1], '"');
```
noplanman: Oh, just saw these lines now, they can be removed.
`preg_match` will fetch exactly what's in the `(.*)`, which is the version without any quotes, so there are no quotes to replace.

Am I missing a special condition that I haven't though of?

(same applies for the other softwares.)

dmorley: Pushed without, I think that was just old code

db/pull.php
```diff
@@ -12,42 +12,42 @@
...
-curl_setopt($mv, CURLOPT_POST, 0);
-curl_setopt($mv, CURLOPT_HEADER, 0);
 curl_setopt($mv, CURLOPT_CONNECTTIMEOUT, 5);
 curl_setopt($mv, CURLOPT_RETURNTRANSFER, 1);
-curl_setopt($mv, CURLOPT_NOBODY, 0);
 $outputmv = curl_exec($mv);
 curl_close($mv);
-preg_match('/number: "(.*?)"/', $outputmv, $version);
-$dmasterversion = trim($version[1], '"');
+$dmasterversion = preg_match('/number:.*"(.*)"/', $outputmv, $version) ? $version[1] : '';
```
noplanman: Awesome use of the ternary operator, nice! 🎉

PR #92, opened 2017-01-29
WIP - Nicer domain checks

noplanman:
Explicitly check passed domain name.

@dmorley I guess it would make sense to use something like this for all places that need a passed domain, what do you think?
Also, since this gives us the IPv4 already, it wouldn't need to be repeated later on (hence WIP). Do you prefer to make the specific checks further down anyway?


noplanman:
> But this would not stop someone from adding diASp.org would it?

Does now :wink:

> That would be nicer for my AAAA check in pull also, mine is a total hack.

This is for IPv4, but there should be something around for IPv6 I guess, will explore.

noplanman:
Erm, how in the world have we not found [this](https://secure.php.net/manual/en/function.dns-get-record.php) yet?! 😕

dmorley:
I remember messing with that, I forgot why I hated it :smile_cat:

noplanman:
> I forgot why I hated it

Well, remember! Cause it does seem to be exactly what we're looking for in this case 👍

dmorley:
I say just do it then!

noplanman:
Ok! And if it gets weird and I start to hate it too, I'll make a note of it 😜

noplanman:
@dmorley 
This should be ok for the moment.

dmorley:
I should just merge that and test the delv, i have a few pods with bogus dnssec records

PR #93, opened 2017-01-29
few misses

dmorley:


showfull.php
```diff
@@ -97,7 +97,7 @@
     $row['service_wordpress'] === 't' && print '<div class="smlogo smlogo-wordpress" title="Publish to WordPress"></div>';
     $row['service_xmpp'] === 't' && print '<div class="smlogo smlogo-xmpp"><img src="/images/icon-xmpp.png" width="16" height="16" title="XMPP chat server" alt="XMPP chat server"></div>';
     echo '</td>';
-    echo '<td><a href="https://' . $row['domain'] . $row['terms'] . '">&#128279;</a></td></tr>';
+    $row['terms'] === 't' ? print '<td><a href="https://' . $row['domain'] . $row['terms'] . '">&#128279;</a></td></tr>' : print '<td></td></tr>';
```
noplanman: This needs to be updated to
```php
    echo '<td>' . ($row['terms'] ? '<a href="https://' . $row['domain'] . $row['terms'] . '">&#128279;</a>' : '&nbsp;') . '</td>';
```
as the field is no boolean.

While we're at it, *must* the terms be on the same domain? I'm thinking of an example where a community might have a general terms page for all products and the pod is running on a subdomain.
In that case, the terms won't be a sub-page of the pod.

db/pull-masterversions.php
```diff
@@ -8,7 +8,7 @@
 $softwares = [
   'diaspora'  => ['url' => 'https://raw.githubusercontent.com/diaspora/diaspora/master/config/defaults.yml', 'regex' => '/number:.*"(.*)"/'],
   'friendica' => ['url' => 'https://raw.githubusercontent.com/friendica/friendica/master/boot.php', 'regex' => '/define.*\'FRIENDICA_VERSION\'.*\'(.*)\'/'],
-  'hubzilla'  => ['url' => 'https://raw.githubusercontent.com/redmatrix/hubzilla/master/boot.php', 'regex' => '/define.*\'STD_VERSION\'.*\'(.*)\'/'],
```
noplanman: Look [here](https://github.com/redmatrix/redmatrix/blob/master/README.md)

Seems to be renamed from redmatrix to hubzilla, so it should stay hubzilla.

dmorley: But the pods nodeinfo/1.0 returns software=redmatrix somewhere along the line that means a hack for now, since the pods are autogenerating and this is manual it made sense to hack here for now.

noplanman: Ok I see. Well in that case leave it till it's "official".

showfull.php
dmorley: Yea I suppose could be anywhere. I wonder if rather than terms, it should be a text field, some limit, 400 or something for a podmin to say something, including links to terms or PP or whatever.

showfull.php
```diff
@@ -97,7 +97,7 @@
     $row['service_wordpress'] === 't' && print '<div class="smlogo smlogo-wordpress" title="Publish to WordPress"></div>';
     $row['service_xmpp'] === 't' && print '<div class="smlogo smlogo-xmpp"><img src="/images/icon-xmpp.png" width="16" height="16" title="XMPP chat server" alt="XMPP chat server"></div>';
     echo '</td>';
-    echo '<td><a href="https://' . $row['domain'] . $row['terms'] . '">&#128279;</a></td></tr>';
+    echo '<td>' . ($row['terms'] ? '<a href="https://' . $row['domain'] . $row['terms'] . '">&#128279;</a>' : '&nbsp;') . '</td>';
```
noplanman: This also needs the `</tr>` to end the row.

PR #94, opened 2017-01-31
Moarfix

dmorley:
Few cleanup items, noticing dead pods getting data from other loops.


db/pull.php
```diff
@@ -97,7 +98,7 @@
     $service_twitter       = in_array('twitter', $jsonssl->services->outbound, true);
     $service_tumblr        = in_array('tumblr', $jsonssl->services->outbound, true);
     $service_wordpress     = in_array('wordpress', $jsonssl->services->outbound, true);
-    $service_xmpp          = $jsonssl->metadata->xmppChat === true ?? false;
+    $service_xmpp          = ($jsonssl->metadata->xmppChat === true) ?? false;
```
noplanman: Out of curiosity, why did you add the braces here? They shouldn't be necessary.

db/pull.php
```diff
@@ -114,20 +115,24 @@
...
+  $dnssec = in_array('; fully validated', $iplookupv4, true) ?? false;
+  $getaonly = array_values(preg_grep('/\s+IN\s+A\s+.*/', $iplookupv4));
   if ($iplookupv4) {
-    _debug('Iplookupv4', $iplookupv4, true);
-    $dnssec = in_array('; fully validated', $iplookupv4, true) ?? false;
-    $getaonly = array_values(preg_grep('/\s+IN\s+A\s+.*/', $iplookupv4));
     preg_match('/A\s(.*)/', $getaonly[0], $aversion);
-    $ip   = trim($aversion[1]);
+    $ip   = trim($aversion[1]) ?? '';
+  } else {
```
noplanman: Instead of this `else`, it's better practice to define the `$ip` variable before the `if`, like so:
```php
$ip = '';
if ($iplookupv4) {
    preg_match('/A\s(.*)/', $getaonly[0], $aversion);
    $ip = trim($aversion[1]) ?? '';
}
```

(same for `$ipv6` below)

showfull.php
```diff
@@ -70,15 +70,15 @@
     }
     echo '<td class="' . $classver . '"><div title="' . $pre . ' version: ' . $row['shortversion'] . ' master version is: ' . $row['masterversion'] . '" data-toggle="tooltip" data-placement="bottom">' . $version . '</div></td>';
     echo '<td>' . $row['softwarename'] . '</td>';
-    echo '<td>' . $row['uptime_alltime'] . '%</td>';
+    echo '<td>' . ($row['uptime_alltime'] > 0 ? $row['uptime_alltime'].'%' : '') . '</td>';
```
noplanman: [Some reading](https://stackoverflow.com/a/2506395) for you.

Not really sure this is super vital either way as I don't know all implications, but still good to know...

db/pull.php
dmorley: I keep getting an error when this metadata does not exist in the nodeinfo at all, I guess I thought I was fixing that!

noplanman: Ok, that's what I thought.
This would be a way to fix it, but can seem a bit confusing:
```php
$service_xmpp          = ($jsonssl->metadata->xmppChat ?? false) === true;
```

PR #97, opened 2017-02-02
Basic pod stats

dmorley:
Lil extra in my SQL query still, but that took me hours to get to work, so leaving for now. Also once 2018 hits this won't work, I need to add "this year" or last 12 months somehow. 

Not sure what all to show here, without much data it looks really boring, I will need to make a years worth of fake data on dev to test properly.  

Don't love blending ttl (int) and uptime (percent) on the same chart, but also don't love just uptime on its own chart. 

I just want to get something basic going here for now so we are closer to a release and can get this on prod and start gathering stats.


dmorley:
@noplanman How is your SQLing? I am really stuck here getting the SQL to be the last 12 months or something, I can do it with PHP but I wanted a clean SQL pull.

noplanman:
> getting the SQL to be the last 12 months or something

Do you mean the last 12 months or the previous year (set of 12 months from the same year)?

Seems to be working for me, or I haven't totally understood what you'd like.
Please explain a bit more, maybe a simple example with some dummy values?

dmorley:
Sorry i fixed since that comment, see last commit

noplanman:
Right, ok.

Will make a PR to your branch to make the SQL easier to read.

Also, I find the lines on the table a bit illogical to look at.
i.e. Having uptime (which is maximum 100) together with latency (which can go way high) on the same graph makes the uptime seem like a straight line at the very bottom of the graph.
Same applies to some of the values on the second graph.

dmorley:
Yea the whole thing is hard to read, maybe more graphs is just better, like split into 3-4?

noplanman:
Maybe, yes. Will have a think...

dmorley:
screenshot from 2017-02-23 11-53-24
I added some fake stats to see how things would graph (just the top one) also, its just "ok" looking to me.

noplanman:
Is it possible to add a second Y scale on the right?

dmorley:
YES! Damn never even crossed my brain

noplanman:
Ok, give that a try. At least it will split the graph up nicer (hopefully).

dmorley:
Yea I think that looks easy to read. Not sure on colors, seems too much. Should it all be tones of the same color per chart or something? No good with design stuff, all black is fine lol

podstat.php
```diff
@@ -0,0 +1,172 @@
...
+      count(autoclick) AS autoclicks
+    FROM clicks
+    WHERE checks.domain = clicks.domain
+      AND EXTRACT(MONTH FROM checks.date_checked) = EXTRACT(MONTH FROM clicks.date_clicked)
+      AND EXTRACT(YEAR FROM checks.date_checked) = EXTRACT(YEAR FROM clicks.date_clicked)
+  ) AS fpp ON TRUE
+  WHERE domain = $1
+  GROUP BY yymm, fpp.manualclicks, fpp.autoclicks
+  ORDER BY yymm
+  LIMIT 24
```
noplanman: Just wondering what this 24 is for. If it's grouped by year + month of the same year, there can never be more than 12, right? So this limit is superfluous, no?

podstat.php
```diff
@@ -0,0 +1,172 @@
...
+";
+$result = pg_query_params($dbh, $sql, [$_domain]);
+$result || die('Error in SQL query: ' . pg_last_error());
+$totals = pg_fetch_all($result);
+?>
+<canvas id="pod_chart1"></canvas>
```
noplanman: I'd suggest better names than just "1" and "2".
There's no length limit here, so just a liiiittle bit more detailed would be cool 😎

showfull.php
```diff
@@ -79,9 +79,8 @@
     echo '<td>' . ($row['active_users_monthly'] > 0 ? $row['active_users_monthly'] : '') . '</td>';
     echo '<td>' . ($row['local_posts'] > 0 ? $row['local_posts'] : '') . '</td>';
     echo '<td>' . ($row['comment_counts'] > 0 ? $row['comment_counts'] : '') . '</td>';
-    $moreurl = '/showstats.php?domain=' . $row['domain'];
     echo '<td><div title="Last Check ' . $row['date_updated'] . '" data-toggle="tooltip" data-placement="bottom">' . $row['monthsmonitored'] . '</div></td>';
-    echo '<td><a rel="facebox" href="rate.php?domain=' . $row['domain'] . '">' . $row['userrating'];
+    echo '<td><a href="rate.php?domain=' . $row['domain'] . '" rel="facebox">' . $row['userrating'];
```
noplanman: Out of curiosity, why move the `rel` here but not in the changed row above?

dmorley: Just bad cleanup from trying a new facebox replacement that did not work. Ill fix.

podstat.php
dmorley: Not sure where your 24 comment is on Limit now, was just thinking at some point 5 years of stats is too much? Am I thinking of the limit wrong?

noplanman: I deleted the comment cause it was wrong. Thought the SQL query limits to 1 year anyway, making the 24 unnecessary, but that's obviously not the case...
So 24 is totally fine as a limit 👌

dmorley:
I'll pad diasp.org with some historical fake data on dev also

PR #98, opened 2017-02-02
some responsiveness cleanup & few features

dmorley:
Few items for cleanup after viewing on a smaller laptop


noplanman:
Might make sense to move [these lines](https://github.com/diasporg/Poduptime/blob/d1d7acea0278981be2029f83d4501a80f6aca125/css/newstyle.css#L62) into this PR instead :+1:

PR #100, opened 2017-02-10
make this more logical and work for pods

dmorley:
Good pods that are now dead are still rated high on dev, and are not loosing any score since the check stops on failure, that only works uphill not downhill. My continue loop was off so I revamped this so even a dead pod will update to defaults and fall off the list.


db/pull.php
```diff
@@ -74,62 +73,62 @@
...
+  $signup                = ($jsonssl->openRegistrations ?? false) === true;
+  $softwarename          = $jsonssl->software->name ?? 'unknown';
+  $name                  = $jsonssl->metadata->nodeName ?? 'null';
+  $total_users           = $jsonssl->usage->users->total ?? 0;
+  $active_users_halfyear = $jsonssl->usage->users->activeHalfyear ?? 0;
+  $active_users_monthly  = $jsonssl->usage->users->activeMonth ?? 0;
+  $local_posts           = $jsonssl->usage->localPosts ?? 0;
+  $comment_counts        = $jsonssl->usage->localComments ?? 0;
+  $service_xmpp          = ($jsonssl->metadata->xmppChat ?? false) === true;
+  $service_facebook      = 'false';
```
noplanman: Use bools here.

db/pull.php
```diff
@@ -74,62 +73,62 @@
...
     $service_twitter       = in_array('twitter', $jsonssl->services->outbound, true);
     $service_tumblr        = in_array('tumblr', $jsonssl->services->outbound, true);
     $service_wordpress     = in_array('wordpress', $jsonssl->services->outbound, true);
-    $service_xmpp          = ($jsonssl->metadata->xmppChat ?? false) === true;
-    $status                = 'Up';
+  }
     
+  if ($jsonssl !== null) {
+    $status        = 'Up';
     $sql_checks    = 'INSERT INTO checks (domain, online, ttl, total_users, local_posts, comment_counts, shortversion) VALUES ($1, $2, $3, $4, $5, $6, $7)';
```
noplanman: I meant to mention this a while back but forgot. The `ttl` here is the connection time in ms, right?
Maybe this can be renamed to prevent confusion with TTL (Time To Live)? This has puzzled me a few times.

dmorley:
Yea not sure why I used that, I guess short name in the table lol. Ping time is still not quite right. connection_time is so long. Any short ideas for a table header word?

dmorley:
Latency is probably the most correct term, is that common? I never use the word.

noplanman:
I'd go with latency. Even though it might not be used often, it sure is the [correct word](https://en.wikipedia.org/wiki/Latency_%28engineering%29) (in my view too).

dmorley:
Not sure what the meaning of dismiss review is, had to push the buttons

db/pull.php
```diff
@@ -86,10 +86,10 @@
   $local_posts           = $jsonssl->usage->localPosts ?? 0;
   $comment_counts        = $jsonssl->usage->localComments ?? 0;
   $service_xmpp          = ($jsonssl->metadata->xmppChat ?? false) === true;
-  $service_facebook      = false;
```
noplanman: Erm, why the int instead of bool?
This is more for readability and not getting confused if it must be bool or int or what.

When inserting into the SQL query, it gets converted to an int anyway, which is what makes most sense I think.

dmorley: lol no idea, I fix it then broke it, no fixed it again

noplanman:
Merge at your leisure 😃

dmorley:
looking good, https://dev.podupti.me/?detailedview but I am unsure about not displaying the 0.00% for uptime, the dead pods are showing a blank there, does that say to the end user: "I don't know" OR "Zero" OR "Never connected to this pod to make a valid calculation". I think I was trying to say the last, but I think a blank in this column might be wrong to the end user. What do you think?

noplanman:
Maybe "TBA" (To be announced) or just a "-"?
I agree with you that 0 or anything "zero" is confusing.

How should dead pods be handled anyway? Maybe have a visual difference? Even keep them in the DB?

dmorley:
"-" looks nice and right. 

I like to give dead pods time to recover, however now that that I changed the score to 100 pt scale you now have to fail over 100 hours/checks to get to zero, that's a lot.  

I was deleting pods manually just to keep the pull.php to a decent run time, it was hitting 30 minutes at one point. But I regretted loosing some of that data. 

I think maybe a deleted flag in the db would be better, the pull could skip deleted=t or run them once a week or something. I guess I think dead pods is an interesting statistic to have over time, maybe only for me?

noplanman:
Deleted flag sounds like a plan, having the statistics is cool 👍

PR #109, opened 2017-02-14
revert curl time for latency value

dmorley:
I was wrong to move to TOTAL_TIME, it seemed logical but as you watch the more users on a pod the longer nodeinfo takes to compile, it's not cached or anything, so a pod can have more users and be shown as 7 seconds, really not that bad, but I think 7000ms will scare away users since our brains are looking for 30-100ms as good latency time on connections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,14 @@ touch add.log in location you configured in config.php
 
 run db/pull.sh manually or with cron to update your data
 
+To Upgrade:
+```
+cd Poduptime
+git pull
+bower install
+psql -u podupuser podupdb < db/migrationx.sql (see db/version.md for proper migration version)
+```
+
 ============================
 
 Source for Diaspora Pod Uptime
--- a/db/pull.php
+++ b/db/pull.php
@@ -63,7 +63,7 @@ while ($row = pg_fetch_assoc($result)) {
   $outputssl      = curl_exec($chss);
   $outputsslerror = curl_error($chss);
   $info           = curl_getinfo($chss, CURLINFO_CERTINFO);
-  $latency        = curl_getinfo($chss, CURLINFO_TOTAL_TIME);
+  $latency        = curl_getinfo($chss, CURLINFO_CONNECT_TIME);
   $sslexpire      = $info[0]['Expire date'] ?? null;
   curl_close($chss);
 
--- a/db/pull.php
+++ b/db/pull.php
@@ -63,13 +63,17 @@ while ($row = pg_fetch_assoc($result)) {
   $outputssl      = curl_exec($chss);
   $outputsslerror = curl_error($chss);
   $info           = curl_getinfo($chss, CURLINFO_CERTINFO);
-  $latency        = curl_getinfo($chss, CURLINFO_CONNECT_TIME);
+  $conntime       = curl_getinfo($chss, CURLINFO_CONNECT_TIME);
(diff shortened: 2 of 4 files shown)


dmorley:
Does your local ever get some outliers latency results? I keep getting some and it's bugging me, over time they would not change the average much but I want to be sure it is not a bug:

podupdb=> SELECT date_checked, round(avg(latency) * 1000) AS latency FROM checks WHERE domain = 'lanx.fr' GROUP BY date_checked;
        date_checked        | latency 
----------------------------+---------
 2017-02-14 14:21:08.315416 |     192
 2017-02-14 15:52:48.66097  |    4681
 2017-02-14 14:20:58.021462 |     161
 2017-02-14 14:20:53.539303 |     183
 2017-02-14 14:21:01.560098 |     164
 2017-02-14 14:36:44.936555 |     165
 2017-02-14 14:21:06.596443 |     161
 2017-02-14 14:20:56.151794 |     160
 2017-02-14 14:21:09.92236  |     160
 2017-02-14 14:21:03.239528 |     161
 2017-02-14 15:02:01.90156  |     160
 2017-02-14 14:21:04.906974 |     164
 2017-02-14 14:20:59.837914 |     162
(13 rows)

podupdb=> SELECT date_checked, round(avg(latency) * 1000) AS latency FROM checks WHERE domain = 'cats.pm' GROUP BY date_checked;
        date_checked        | latency 
----------------------------+---------
 2017-02-14 14:50:44.69495  |     167
 2017-02-14 14:45:34.124185 |     172
 2017-02-14 14:56:51.080067 |     164
 2017-02-14 14:40:44.07878  |     169
 2017-02-14 14:45:30.232104 |     165
 2017-02-14 14:30:55.330129 |    4684
 2017-02-14 15:47:28.065348 |     672
 2017-02-14 14:45:32.176238 |     164
 2017-02-14 14:40:53.851742 |     167
 2017-02-14 16:00:15.714337 |     170
(10 rows)

just 2 pods but most have a random one way off, and not from the same run so its not like all get slow on a run, that would be logical. I noticed this on dev also, some really long times when most are not at all. Could be a million things, could be when my dns cache is expired and it has to lookup the IP again, could be other processes doing things... just no consistency.

dmorley:
Sorry should have included that I can not get this result when running curl cli or when running db/pull.php with one domain. IT is only when running db/pull.php for all domains that this happens. So my brain says some bug is in there when this updates all.

noplanman:
This is obviously not from the recent change when you set it to `CURLINFO_TOTAL_TIME`, right?

If not, then you pretty much nailed it with:

> Could be a million things

Alternatively, you could make the request 3 times or so, taking just the lowest result. This would (kind of) make sure that rare slow connection times don't count.

dmorley:
Yea this is using the conn time only. Flushed db table to be sure I wasn't 
getting mixed.

Hrm ill keep testing a bit.

noplanman:
I mean there's always a chance that a connection may take a while, which is also the reason why speedtests should be done multiple times, not just once.

Not 100% sure, but I don't think this is due to a bug.

Maybe in the future we move to use [Guzzle](https://github.com/guzzle/guzzle) for all HTTP handling, which should optimise it even further.

dmorley:
it was dns lookup time that was the issue. That is a user issue not a host issue. So the answer is conn time minus ns time is latency. The numbers now actually tie with when you click a high number it connects slow and a low number is fast, magic.

PR #115, opened 2017-02-20
SQL Reloaded!

noplanman:
New DB management using doctrine's DBAL library.

This requires `composer install` to be called first.

@dmorley Still WIP, but I thought I'd open already to let you have a look at the progress :wink:


dmorley:
Finally running this local, looks great, it's pretty easy to understand really. Still looking over it, but it all works!
Don't forget to add composer to the readme as a core package you need in your OS. Can composer replace bower for things? Would be nice to just have one of them.

noplanman:
> Don't forget to add composer to the readme

Yep. Will update the readme as the last step 👍

db/pull.php
```diff
@@ -4,42 +4,72 @@
...
-  $ratings || die('Error in SQL query2: ' . pg_last_error());
+
+try {
+  $sql = $db_conn->createQueryBuilder()
+    ->select('domain', 'score', 'date_created', 'adminrating', 'weight', 'hidden', 'podmin_notify', 'email')
+    ->from('pods');
+
+  $pods = [];
+  if ($_domain) {
+    $sql->where('domain = ?');
```
dmorley: I think I am missing how ? gets defined

noplanman: The very next line `[$_domain]`. The parameter/array after that, says what data type is to be used. Not 100% sure yet if this it *totally* necessary, will need to learn more about that.

This is the same as the [`pg_query_params`](https://secure.php.net/manual/en/function.pg-query-params.php)

api.php
```diff
@@ -21,71 +36,110 @@
...
+ * Convert all passed items to strings.
+ * 
+ * This method is for backwards compatibility of APIv1 only!
+ * After v2 is released and stable, this can safely be removed.
+ *
+ * @param array $arr List of all elements to stringify.
+ *
+ * @return array
+ */
+function allToString(array $arr)
```
noplanman: This function is necessary for backwards compatibility, to make sure the API output is **exactly** the same.
With the new v2, this can be deleted, as types will be cleaner then, letting integers and booleans actually be their own types, not everything as string.

db/api-more.php
```diff
@@ -8,27 +8,50 @@
...
+  die('Error in connection: ' . $e->getMessage());
+}
+
+try {
+  $sql  = $db_conn->createQueryBuilder()
+    ->select('hgitdate', 'id', 'domain', 'status', 'secure', 'score', 'userrating', 'adminrating', 'city', 'state', 'country', 'lat', 'long', 'ip', 'ipv6', 'pingdomurl', 'monthsmonitored', 'uptimelast7', 'responsetimelast7', 'local_posts', 'comment_counts', 'dateCreated', 'dateUpdated', 'dateLaststats', 'hidden')
+    ->from('pods_apiv1')
+    ->where('domain = ?');
+  // For backwards compatibility, fetch associative and indexed values.
+  $db_conn->setFetchMode(PDO::FETCH_BOTH);
```
noplanman: @dmorley This is to make sure the output is exactly the same, but I'm not sure this is actually necessary.
I think a simple associative array is enough here, don't think the indexes are needed.

i.e.
```json
{"id":1,"0":1,"domain":"diasp.org","1":"diasp.org","status":"Up","2":"Up"}
```

```json
{"id":1,"domain":"diasp.org","status":"Up"}
```

noplanman:
This PR may introduce more code for now, but as soon as we move to a simpler framework, most of the code can be removed.

dmorley:
Been running this a few days in my local dev, looking good, no errors and it makes sense.

noplanman:
@dmorley Might have found something better. If not, I'll reopen this.

PR #117, opened 2017-03-01
fix hangs

dmorley:
I don't love a *nix only command like timeout but I think delv already is. All 3 of my environments including prod are hanging on some dns calls with delv and I have to go and manually kill the delv run off.


noplanman:
Have you come across and tried this?
https://stackoverflow.com/a/16407770

dmorley:
I did, I tried it with popen and proc_open and with exec and was not successful at getting it to work with exec and not successful getting delv to work with proc_open. I searched for better answers and came up with nothing around proc_* and no path I could find.

noplanman:
Ok. I'll have a quick look anyway when I get the chance to, maybe there is some other solution out there somewhere.

noplanman:
Right, I have a solution for this. After checking various libraries that already exist for command execution via PHP, none of them support timeout.

So, what does a programmer do? Exactly, makes one!

It's still not totally ready, but will release it soon 👍

dmorley:
Sweet! Super weird timeout is not a universal thing!

noplanman:
Whoops, just realised I pushed my commit to your PR directly, wasn't the idea, but since it's already in here, take a look 👀

dmorley:
Nice will pull and test. You should add more description to your xec readme 
since your solving world issues here, nobel prize incoming (that's a cookie 
from me, best I can do)

noplanman:
Haha, yes, the readme is still a bit empty... Also, I'll be changing the project quite a bit to make it more logical to execute commands, similar to https://github.com/demvsystems/exec

That project does have a way to execute a command asynchronously, but then there is no way to get the actual output of the command, which is a real pity.

dmorley:
Been running this 24 hours on my local dev, working good, lets merge and push to dev and see how it goes, dev has been hanging daily so good test.

noplanman:
Anything special that needs to be done when delv times out?
Maybe 3 retries or something?
If not, you can merge this.

dmorley:
I don't think so, since it will just run again in an hour

PR #118, opened 2017-03-07
Use Redbean as DBAL

noplanman:
Use RedBean as a database abstraction layer, making the whole DB queries so much easier.

This is still WIP, because some queries don't work properly with the existing database structure, which I'm still trying to figure out. So far this is only a problem for `pull.php`, when updating the pod record at the end of the script.

@dmorley Have a look at this and let me know if this is similarly easy to understand as the doctrine DBAL PR I closed the other day. This PR has all the commits of that branch in it too, with 1 extra one changing the DBAL to RedBean, just in case.


dmorley:
This is actually easier to read to me because the SQL is still SQL. I wonder what is the overall gain here then? Not objecting at all, just trying to understand RB over plain ole pgcon

noplanman:
Read the [quick tour](http://redbeanphp.com/index.php?p=/quick_tour) and you'll see the huge advantage!

As for us at this stage, the advantage is, that using a library that has been tried and tested is much more reliable than homebrewing something. Also, the awesome thing about RedBean is what I'm doing [with `$pod` here](https://github.com/diasporg/Poduptime/pull/118/files#diff-1841ae349cbde0fddc107420293356d2R27).

If you look further down in the code, it's super easy to update an already fetched row, or delete it with a single call, no SQL required. To me, it makes it a LOT more readable and less prone to error.

noplanman:
@dmorley Ok, this should be good to go now, had to add a fix which seems to be a bug in the RedBean library.

dmorley:
Since db/migration00001.sql already is in master shouldn't this be db/migration00002.sql now just adding your ID's? Since I already have the tables from 001 I would have to drop them to do the 001 migration now.

noplanman:
D'oh, of course... Will fix this.

dmorley:
Does work well however on my dev, I just modified the tables as needed.

noplanman:
@dmorley Have added the dedicated migration script, which adds the unique primary key column `id` and sets all previously existing rows correctly.

dmorley:
Just want to make sure I get it:
On create of a table:
id serial8 UNIQUE PRIMARY KEY

but on migrate:
CREATE SEQUENCE apikeys_id_seq;
ALTER TABLE apikeys ADD id serial8 UNIQUE PRIMARY KEY;
ALTER TABLE apikeys ALTER COLUMN id SET DEFAULT NEXTVAL('apikeys_id_seq');
UPDATE apikeys SET id = NEXTVAL('apikeys_id_seq');

So if you create a serial8 UNIQUE PRIMARY KEY it already knows to use a sequence but on an alter you need to be more manual about it?

I of course just something basic when I was testing and did:
alter table apikeys add id serial8 UNIQUE PRIMARY KEY

noplanman:
The manual work around is necessary for a migration, when the table already has rows in it.
Because a new PRIMARY KEY column is being added, the existing rows must all have a unique entry in it, which by default is just NULL (causing an error).
For this reason, a new SEQUENCE must be introduced for each table and then all existing rows must be set to use that SEQUENCE, in essence filling the rows with sequential unique IDs.

If there's an easier way to do this, I'm all ears 👂

Will fix the conflicts tomorrow!

dmorley:
Nope it makes sense now, I was just not connecting the dots.

dmorley:
updated dev and I get an error when I try and edit an call to gettoken.php

Error in SQL query: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type boolean: ""

I think that has to be line 53 or 54 but neither are boolean

noplanman:
Check #119

dmorley:
Found one more thing (sorry I clearly did not test great)

in edit.php

<?php $pod['podmin_notify'] === 't' ?? 'CHECKED' ?> on line 80

So since redbean brings back podmin_notify as bool this can just be 

<?php $pod['podmin_notify'] ?? 'CHECKED' ?> now like I did it originally as I assumed bool was bool and now it really is?

noplanman:
Yes! Absolutely.

I could have sworn replaced all of them, hmm... 😕

noplanman:
Wait a sec, this should be:
```php
<?php echo $pod['podmin_notify'] ? 'checked' : ''; ?>
```

dmorley:
Yes. And weird that its never had the ; at the end but no error? Seems like that should break php no?

noplanman:
Nope, PHP will allow this if it's the last command within a `<?php ?>` block.
```sh
$ php <<< '<?php echo "a"; echo "b" ?>
```
works perfectly.

dmorley:
I think your hack messes with number fields, uptime_alltime, latency are null now on dev. Trying to debug myself but no progress yet.

dmorley:
Only progress I made is if you call a single pod via cli it works, if you use db/pull.sh it does not, so something just a bit wonky somewhere.

dmorley:
I think same bug on c insert, so I tried:
   `foreach ($c->getProperties() as $key => $value) {
      $pod[$key] = $value;
    }`
like your other hacks with no glory, I don't really understand the bug that well.

noplanman:
Did you read the issue I posted on the RedBeans repo?
In your example, you forgot to change `$pod[$key]` to `$c[$key]`.

Time to get some unit testing done, this is getting out of hand 😨

dmorley:
I read it, I need to read it again for sure.

I copied you on the bottom of pull.php and you used 

`foreach ($p->getProperties() as $key => $value) {
      $pod[$key] = $value;
    }`

So should your $pod[$key] be $p[$key] or am I totally lost at this point?

noplanman:
Yes it should have been 😦 
Damn how are these things slipping through, so sorry!

dmorley:
No worries, its why we try stuff out. However I fixed that and the 2 others and no glory, I get no errors but my table is a bore. 
screenshot from 2017-03-17 09-43-33

noplanman:
Ok, will look into that right now ❗️

noplanman:
Right, I don't experience this issue on my dev...
I've loaded the most up-to-date DB that I have from you and pulled a few pods, all work totally fine.

Have you tried with a fresh dump or does the pull literally kill all the values?
Also, are they all NULL in the DB then?

dmorley:
I get success on pulling one at a time via cli or webserver. I get the issues when I run the pull.sh for all pods. Try a pull.sh run?

dmorley:
And I just checked in the dev and my local db and the value is 0.00 



```
dev=> select uptime_alltime from pods where domain = 'diasp.org';
 uptime_alltime 
----------------
           0.00
```

PR #119, opened 2017-03-16
Temporary fix for pods table boolean fields.

noplanman:
Fix the issue when storing rows that might have "unset" booleans.

Really hope this gets fixed in RedBean at some point...


dmorley:
Ah yea I forgot about that, I will watch the bug you opened also and see where it goes.

PR #123, opened 2017-03-17
Fix wrong variable name and correct RedBeans fix.

noplanman:


db/pull.php
```diff
@@ -211,7 +211,7 @@
         round(avg(online::INT) * 100, 2) AS online
       FROM checks
       WHERE domain = ?
-    ', [$_domain]);
+    ', [$domain]);
```
noplanman: This was the culprit, setting all checks to something or other, instead of just the *current* domain.

dmorley: bam !

PR #126, opened 2017-03-18
index

dmorley:
The LEFT JOIN LATERAL is damn slow on the clicks on prod, an index makes it right for now. I think as that table grows we might have to pull monthly stats to its own table and not pull as requested.


noplanman:
The classical things that work great on dev but then start showing in prod :wink:

dmorley:
Yea and I tested with the checks and pods table on local. But clicks seemed 
a bit silly since its a low #, turns out on prod that's not true, I'd guess 
a lot of bot clicks also in there i need to poke around and see and filter out.

dmorley:
Yes, 95% spiders in the clicks table when looking at nginx log. So here is a fix for that also as part of this PR to deal with slow query speed there.

go.php
```diff
@@ -36,7 +39,7 @@
   $c           = R::dispense('clicks');
   $c['domain'] = $domain;
   $c[$click]   = 1;
-  R::store($c);
+  if(!$CrawlerDetect->isCrawler()) {R::store($c);}
```
noplanman: For prettification sake:
```php
if (!(new CrawlerDetect())->isCrawler()) {
  R::store($c);
}
```
Then you can remove the assignment on line 17, as you're not interested in doing anything with the object anyway.

noplanman:
They won't be, as composer is purely for PHP things, whereas bower is for javascript.

PR #127, opened 2017-03-19
WIP better delete

dmorley:
I will want to write tests first on this. 

I am thinking about killing 'hidden' from pods table and making that a status state, thoughts? Only reason I did not do it was making it work for the old api but that should not stop progress.

### DIFF
--- a/db/api-more.php
+++ b/db/api-more.php
@@ -18,6 +18,7 @@ define('PODUPTIME', microtime(true));
 // Set up global DB connection.
 R::setup("pgsql:host={$pghost};dbname={$pgdb}", $pguser, $pgpass, true);
 R::testConnection() || die('Error in DB connection');
+R::usePartialBeans(true);
 
 try {
   $pod = R::getRow('
@@ -32,10 +33,10 @@ try {
 if ($_format === 'json') {
   echo json_encode($pod);
 } else {
-  if ($pod['status'] === 1) {
+  if ($pod['status'] === PodStatus::Up) {
     echo 'Status: Up<br>';
   }
-  if ($pod['status'] === 0) {
+  if ($pod['status'] === PodStatus::Down) {
     echo 'Status: Down<br>';
   }
   echo 'Last Git Pull: ' . $pod['hgitdate'] . '<br>';
--- a/db/api-more.php
+++ b/db/api-more.php
@@ -32,7 +32,12 @@ try {
 if ($_format === 'json') {
   echo json_encode($pod);
 } else {
-  echo 'Status: ' . $pod['status'] . '<br>';
+  if ($pod['status'] === 1) {
+    echo 'Status: Up<br>';
+  }
+  if ($pod['status'] === 0) {
+    echo 'Status: Down<br>';
+  }
   echo 'Last Git Pull: ' . $pod['hgitdate'] . '<br>';
   echo 'Uptime This Month ' . $pod['uptimelast7'] . '<br>';
   echo 'Months Monitored: ' . $pod['monthsmonitored'] . '<br>';
--- a/db/edit.php
+++ b/db/edit.php
@@ -1,6 +1,7 @@
 <?php
 
 use RedBeanPHP\R;
+use CommerceGuys\Enum\AbstractEnum;
 
 // Required parameters.
 ($_domain = $_GET['domain'] ?? null) || die('no pod domain given');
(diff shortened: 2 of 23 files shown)


db/edit.php
```diff
@@ -35,8 +35,47 @@
 
 // Delete and exit.
 if ('delete' === $_action) {
-  R::trash($pod);
-  die('pod removed from DB');
+  try {
+    $pod['status']            = 5;
```
noplanman: I'd highly recommend to go this way, more OOP and a lot easier to manage:
https://stackoverflow.com/a/254543

With an enum, the code is a lot more read, as you're saying what it is, as opposed to some number which isn't clear.

i.e. `$pod['status'] = PodStatus::UP;` instead of `$pod['status'] = 0;`

db/api-more.php
```diff
@@ -32,7 +32,12 @@
 if ($_format === 'json') {
   echo json_encode($pod);
 } else {
-  echo 'Status: ' . $pod['status'] . '<br>';
+  if ($pod['status'] === 1) {
```
noplanman: Using the approach linked below, this can be made a lot easier (pseudocode, just to illustrate):
```php
echo 'Status: ' . PodStatus::getDescription($pod['status'] ?? -1) . '<br>';
```
`getDescription` would return the human readable "value" of the status.

db/edit.php
```diff
@@ -35,8 +35,47 @@
 
 // Delete and exit.
 if ('delete' === $_action) {
```
noplanman: Would make sense to extract this into a function, as (pretty much) each status will have 95% the same code.

db/migration00003.sql
```diff
@@ -1,2 +1,13 @@
...
+UPDATE pods SET status=1 WHERE status = 'Up';
+/*
+0 = Down
+1 = Up
+2 = Recheck
+3 = Paused
+4 = System Deleted
+5 = User Deleted
+*/
+ALTER TABLE pods ALTER COLUMN status TYPE smallint USING status::smallint;
```
noplanman: Just pointing out that enums exist for PostgreSQL too:
https://wiki.postgresql.org/wiki/Enum

Not sure how PHP and RedBeans handles enums, but an int here is totally fine too to start with I think 👍

dmorley:
I was thinking I would add another state for Hidden, Maybe "Auto Pause" and "User Pause". However on pondering more, I really have no good reason to remove the hidden row and it just adds headache around api v1 that might not be needed right now.

db/edit.php
```diff
@@ -108,7 +109,20 @@
...
+    const Down           = 0;
+    const Up             = 1;
+    const Recheck        = 2;
+    const Paused         = 3;
+    const System_Deleted = 4;
+    const User_Deleted   = 5;
+}
+$test = $pod['status'];
+echo $test;
+$status = PodStatus::getKey($test);//value is blank
```
dmorley: Help? This should be easy, no matter what I try assigning the value of $pod['status'] to getKey I fail.

db/gettoken.php
```diff
@@ -26,7 +26,7 @@
 
 // Set up common variables.
 $uuid          = md5(uniqid($_domain, true));
-$link          = sprintf('https://%1$s/db/edit.php?domain=%2$s&token=%3$s', $_SERVER['HTTP_HOST'], $_domain, $uuid);
+$link          = sprintf('https://%1$s/?edit&domain=%2$s&token=%3$s', $_SERVER['HTTP_HOST'], $_domain, $uuid);
```
dmorley: I am missing something basic here. The edit.php is called and the domain and token are validated, but then the page is blank. It should display the forms since I passed the checks right?

db/edit.php
noplanman: Probably because you have a string in `$test`, try: `$test = (int)$pod['status'];`

db/gettoken.php
dmorley: Ah, because the db is already set in index so calling it again is an error. I think putting edit in index looks a lot nicer.

db/edit.php
dmorley: Yea that was it. So just like an array I have to define the enum on each php page, is their some way to define these just one time that all pages can use other than just including?

db/edit.php
dmorley: YES

db/gettoken.php
noplanman: Going towards Slim, everything will be in index! 😃 

And yes, I totally agree that it's nicer.
I held back with just adding everything to index to keep the steps small and manageable, but I'm obviously all for putting them together.

dmorley: yes. Well slim will kill this but was good to do and learn why it was bratty.

dmorley:
@noplanman have any recall of this PR, it is good to go or best to delete if I am going to do some work? I can't remember it at all

noplanman:
Don't remember it in detail either, but checking through the code, it looks almost ok.
I noticed that the pod status enums aren't used yet, there are still integers assigned instead of `PodStatus::XYZ`.

Also, the temporary fix for RedBean can be removed, as it's been fixed a while back:
https://github.com/gabordemooij/redbean/issues/547

noplanman:
@dmorley I'll update this PR, making use of enum and fixing the RedBean stuff, ok?

dmorley:
Should I merge this PR or kill it off now? I think your changes are in place of?

noplanman:
My changes went into this, so this would now have to be merged into develop :+1:

Issue #130, opened 2017-08-25
Remove outdated pods from the list


SuperTux88:
Can you please remove all pods older than 0.6.3.0 from the "Simple view" list (and the list used for random selection)? They aren't compatible anymore with the latest federation protocol. Especially diasp.ca, which is at the top, but has still 0.5.x ... so every user registering there is a lost user.

For long-term it would be cool to remove all pods that aren't updated in the last 6 month (4 minor releases), even when still compatible. But they are missing new features and probably security fixes, so it's nothing we should recommend for new users when the pod isn't well maintained.

SuperTux88:
Can you please remove these outdated pods? Since the release of 0.7.0.0 more than 100 users registered on diasp.ca, they are now trapped on an outdated, unmaintained pod which isn't compatible with the network anymore. And that's only one pod, I didn't check the others. But diasp.ca is the biggest problem, because it is at the top of the list, has 100% uptime and has a lot of users, so it looks interesting, but it is the worst choice.

dmorley:
A programmatic approach would be needed that represents all pods on all networks. Does nodeinfo offer some sort of compatible versions list? Or some way to tell the codebase what versions are compatible or not. How do diaspora nodes know to stop talking to these other nodes that are not compatible anymore?

SuperTux88:
> A programmatic approach would be needed that represents all pods on all networks.

The easiest way is probably to only list the pods that are updated in the last 6 months. Because even if they are still compatible, a pod that wasn't updated for more than half a year is nothing I would recommend to new users (a lot of new stuff is missing and I don't know if the pod is updated in the future or if it stays unmaintained). Diaspora releases every 6 weeks, so that would be "only display the last 4 minor releases" (hotfix releases don't count here).

> How do diaspora nodes know to stop talking to these other nodes that are not compatible anymore?

They don't but the old pods just don't understand the new pods anymore. And even when it's not a 100% incompatible and they still understand the basics, older pods don't understand new stuff that was added since then.

So it is really important that podmins update their pods regularly (also because of security hotfixes).

dmorley:
Only diaspora shows when the code was updated in the http headers so that would not work overall I think. So no easy way to know if someone is updating their code or not that I can think of?

I just had a thought. When did nodeinfo 2.0 come out? If I just change from checking 1.0 to 2.0 does that mostly solve this?

SuperTux88:
[...]

PR #131, opened 2017-09-10
fix for some really old and incompatable pods - this should be done a…

dmorley:
…t some point also using the data from the checks table

### DIFF
--- a/db/pull.php
+++ b/db/pull.php
@@ -228,6 +228,11 @@ foreach ($pods as $pod) {
   }
 
   _debug('Masterversion', $masterversion);
+  $masterversioncheck = explode('.',$masterversion);
+  $shortversioncheck = explode('.',$shortversion);
+  if (($masterversioncheck[1] - $shortversioncheck[1]) > 1) {
+    _debug('Outdated', 'Yes');$score -= 2;
+  }
 
   $hidden = $score <= 70;
   _debug('Hidden', $hidden ? 'yes' : 'no');
--- a/db/pull-masterversions.php
+++ b/db/pull-masterversions.php
@@ -16,7 +16,7 @@ $softwares = [
   'diaspora'     => ['url' => 'https://raw.githubusercontent.com/diaspora/diaspora/master/config/defaults.yml', 'regex' => '/number:.*"(.*)"/'],
   'friendica'    => ['url' => 'https://raw.githubusercontent.com/friendica/friendica/master/boot.php', 'regex' => '/define.*\'FRIENDICA_VERSION\'.*\'(.*)\'/'],
   'redmatrix'    => ['url' => 'https://raw.githubusercontent.com/redmatrix/hubzilla/master/boot.php', 'regex' => '/define.*\'STD_VERSION\'.*\'(.*)\'/'],
-  'socialhome'   => ['url' => 'https://raw.githubusercontent.com/jaywink/socialhome/master/bower.json', 'regex' => '/version":.*"(.*)"/'],
+  'socialhome'   => ['url' => 'https://raw.githubusercontent.com/jaywink/socialhome/master/socialhome/__init__.py', 'regex' => '/__version__ =.*"(.*)"/'],
   'social-relay' => ['url' => 'https://raw.githubusercontent.com/jaywink/social-relay/master/social_relay/config.py', 'regex' => '/VERSION.*"(.*)"/'],
 ];
 


db/pull.php
```diff
@@ -228,6 +228,11 @@
   }
 
   _debug('Masterversion', $masterversion);
+  $masterversioncheck = explode('.',$masterversion);
+  $shortversioncheck = explode('.',$shortversion);
+  if (($masterversioncheck[1] - $shortversioncheck[1]) > 1) {
+    _debug('Outdated', 'Yes');$score -= 2;
```
SuperTux88: The score will be reduced at every check, so after a few checks outdated pods are hidden?

noplanman: -2 enough to kick them off the list?
If they're outdated like that, they should just be hidden completely, even if they get all the other points, right?

dmorley: I think so, at this point they have such old code it's not going to work with the network

dmorley: Yea within a day they would drop off the main list and auto select list

noplanman: This will only work properly for diaspora pods, as the versioning of Friendica, Hubzilla and SocialHome are different. (0.x.y.z vs. x.y.z)

For the others, a minor version difference will lead to being hidden.
As a quick fix for a quick fix, apply this check just for diaspora pods?

dmorley:
dev redirect is well broken on nginx or somewhere. this is testing at https://podupti.me/dev/?detailedview now

db/pull.php
dmorley: Well I took a look at their release schedule
https://github.com/friendica/friendica/releases
https://github.com/redmatrix/hubzilla/releases
https://github.com/jaywink/socialhome/releases

And while it's not perfect and maybe unfair, it seems like if you are > 1 from second decimal then you are not updating enough to keep your code base in line. So it feels ok to me, seems too harsh?

noplanman: Looking at it that way, I guess not.

But then this check does nothing when master is say `3.0.0` and pod is on `2.3.4`. Which would be even more extreme than a minor being 1 off 😕 

Having said that, this is temporary (🙏), so it should be ok, hoping that none of them go for a new major release any time soon.

SuperTux88: I think for friendica it's OK too (at least for now as a quickfix), they have 3.5.2 now and 3.5.3 coming and next will be 3.6.0. But it breaks when there is a 4.0.0. And socialhome has 0.x too, so it will work too. I'm not sure about hubzilla.

If you do a special check only for diaspora can you do something like this?

```php
$masterversioncheck[2] >= 3 ? ($masterversioncheck[1] != $shortversioncheck[1] && ($masterversioncheck[2] - $shortversioncheck[2]) > 3) : (($masterversioncheck[1] - $shortversioncheck[1]) > 1)
```

That implements parts of my comment here https://github.com/diasporg/Poduptime/issues/130#issuecomment-328346421

noplanman: @SuperTux88 Sonderwünsche? 😜

When there is a rewrite, it would be nice to have some criteria when a pod is outdated, which gets submitted by the core devs of each project. That way, they get to decide when a version is deemed outdated. What do you think?

SuperTux88: > @SuperTux88 Sonderwünsche? 😜

hehe ;)

> When there is a rewrite, it would be nice to have some criteria when a pod is outdated

It's difficult to write good criteria for when a pod is outdated, even my "special diaspora check" above isn't perfect (because it removes 0.7.0.0 when 0.7.4.0 is released (which is good), but adds 0.7.0.0 again once 0.8.0.0 is out, because I don't know what's the last minor of 0.7.x was). 

So the simplest and most flexible solution would be to have a maintained list of allowed versions (where new versions are added automatically) which the core devs can maintain and remove old versions if needed.

dmorley:
updated socialhome. I wonder if the right way to check all latest versions is using the github api for release numbers rather than checking inside the code base. That makes the assumption that all devs will use the release functionality

db/pull.php
noplanman: That would be the best, agreed.

So it would mean getting in touch with the core devs of the projects and agreeing on a way to implement this? As part of the node info or config file?

noplanman:
@dmorley I think code base is better, as that (should) be the code that is running. And a release is only a tag of that code anyways.

SuperTux88:
Thanks for merging, is this already deployed? Because it looks like after the last check diasp.ca has still a score of 100.

> I think code base is better, as that (should) be the code that is running.

I disagree, parsing it somewhere out of the code is different for every project and can break when something in the project changes. Using GitHub release API is the same for every project and doesn't break when one project changes something. Diaspora uses the GitHub API too to check if there is an update available in the admin panel (`/admins/dashboard`), and publishing the release on GitHub is part of the release-checklist.

> So it would mean getting in touch with the core devs of the projects and agreeing on a way to implement this? As part of the node info or config file?

Since this would be podupti.me specific I thought about something on podupti.me. Since there is already something like a login system for managing the pod description and stuff, maybe you can create an account for each project and a small page to add/remove versions from a list (doesn't need to be something complex or fancy, a big textarea with one version per line would do the job). NodeInfo doesn't work, because every pod has it's own NodeInfo and there is no NodeInfo for the project and where would you place such config file?

dmorley:
It's on dev now, I will deploy to production after a few more checks today and make sure things run as intended.

PR #133, opened 2018-03-24
some cleanup and fix the stats that just grind

dmorley:
Just fix a few things that look funny and some stats stuff that is super slow right now

### DIFF
--- a/show.php
+++ b/show.php
@@ -39,7 +39,7 @@ try {
     $verdiff  = str_replace('.', '', $pod['masterversion']) - str_replace('.', '', $pod['shortversion']);
     $pod_name = htmlentities($pod['name'], ENT_QUOTES);
     $tip      = sprintf(
-      'Uptime %2$s%% over %1$s months, score is %3$s out of 100.',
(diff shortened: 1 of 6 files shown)


show.php
```diff
@@ -39,7 +39,7 @@
     $verdiff  = str_replace('.', '', $pod['masterversion']) - str_replace('.', '', $pod['shortversion']);
     $pod_name = htmlentities($pod['name'], ENT_QUOTES);
     $tip      = sprintf(
-      'Uptime %2$s%% over %1$s months, score is %3$s out of 100.',
+      'Uptime %2$s%% over %1$s months.',
       $pod['monthsmonitored'],
       $pod['uptime_alltime'],
       $pod['score']
```
noplanman: This line can be deleted.

noplanman: I meant only the line 45 can be removed, as it's not used in the `sprintf` string any more :wink:

dmorley: Got it! Thanks hard to tell on mobile :)

PR #135, opened 2018-03-26
stats_api key should still be in the db

dmorley:
Setting up dev on my laptop again and this item should still be in the db to support the legacy api, but is not really used so removed from the calls.


dmorley:
Because the old api converter tries to convert it back to pingdomkey to keep that api structure alive

noplanman:
Riiiight, gotcha :+1:

PR #136, opened 2018-03-27
users line chart

dmorley:
Not sure I have the SQL quite right. Regardless pulling lots of check data is going to always be a tough pull. I am wondering if a new table should be made with monthly overall stat data so queries like this would be fast. Not sure the right approach to go with now, just know this is too slow. Ideas?


dmorley:
I thought about this more and a pretty simplistic way is to sample one day and just divide it out, the query is quick that way, but I imagine would slow over time. 

    SELECT
      to_char(date_checked, 'yyyy MM') AS yymm,
      sum(total_users) / 24 AS users
    FROM checks
    WHERE to_char(date_checked, 'DD') = '20'
    GROUP BY yymm
    ORDER BY yymm

statsviewjs.php
```diff
@@ -20,11 +20,13 @@
 }
 
 try {
+  R::useWriterCache(true);
```
noplanman: Sweet, was going to suggest caching! Didn't know that RedBean has something of its own.

It might make sense to use a proper cache at some point to not need even a 2 second query :innocent:

statsviewjs.php
```diff
@@ -20,11 +20,13 @@
...
   $check_totals = R::getAll("
     SELECT
-      to_char(date_checked, 'yyyy MM') AS yymm,
-      sum(total_users) / count(DISTINCT domain) AS users
+      to_char(date_checked, 'yy mm') AS yymm,
+      sum(total_users) / 24 AS users
     FROM checks
+    WHERE to_char(date_checked, 'DD') = '20'
```
noplanman: So basically only get the average values for the 20th of each month?

I haven't quite understood the `/ 24` division part, what exactly are you querying here?

dmorley: Yea so doing every day of the month was 18 seconds so why not just a sample day. I pull stats every hour so /24 but I feel like there should be a programmatic way do get a divisible number somehow, nothing working in my brain.

noplanman: How about a temporary table in the DB that simply saves the calculated values for the stats to access?

So after each pull (or even just once a day maybe) do the maths once and save it. Then, when a user checks the stats page, it's as simple as `SELECT yymm, users FROM checks_cache ...`.

dmorley: Yea I think that makes the most sense, it would be nice to pull more than just users on a graph and if this is pushing 3 seconds that is still a slow page load time and a live hit to the db each run. 

Pondering if it should be daily or monthly, I don't have a use case for charting anything by day in my head right now.

noplanman: Regarding the charts, would make sense to not just have the same ones as https://the-federation.info/

Also, getting data from their API from time to time would also make sense to get a more complete list.

dmorley: Yes and No I think, not sure what else to chart, users seems the only thing interesting to the masses. I don't think we need more on a bar for now, but I would set the table up to have more numbers they might be a nice API call one day.

I run the podcrawler on production once a day, https://github.com/diasporg/Poduptime/blob/master/podcrawler.php

db/monthly_stats.php
```diff
@@ -0,0 +1,27 @@
...
+
+use RedBeanPHP\R;
+
+defined('PODUPTIME') || die();
+
+try {
+  $monthly_totals = R::getAll('
+    SELECT
+    
+#select to_char(date_checked, 'mm yy') AS yymm, avg(total_users) as users, avg(local_posts) as posts, avg(comment_counts) as comments, count(DISTINCT domain) as pods, round(avg(online::INT),2)*100 AS uptime from checks WHERE online GROUP by yymm ORDER by yymm;
```
dmorley: @noplanman man this is killing my brain, it should be easier to pull this now that I don't care if it takes a minute. Any smoother ways to pull this off?

noplanman: Will check it out when I get home later today :+1:

db/monthly_stats.php
```diff
@@ -2,26 +2,48 @@
...
+  #format date to timestamp
+  $timestamp = $monthly['yymm'].'-01 01:01:01-01';
+  
+  try {
+    $p                   = R::dispense('monthlystats');
+    $p['total_users']    = $monthly['users'];
+    $p['total_posts']    = $monthly['posts'];
+    $p['total_comments'] = $monthly['comments'];
+    $p['total_pods']     = $monthly['pods'];
+    $p['total_uptime']   = round($monthly['uptime'] / $monthly['downtime'],0);
```
dmorley: I tried and tried to do this inside the SQL query. Maybe you see the light there @noplanman

db/monthly_stats.php
```diff
@@ -0,0 +1,49 @@
...
+  die('Error in SQL query: ' . $e->getMessage());
+}
+
+foreach ($monthly_totals as $monthly) {
+
+  #format date to timestamp
+  $timestamp = $monthly['yymm'].'-01 01:01:01-01';
+  
+  try {
+    $p                   = R::dispense('monthlystats');
```
dmorley: dispense is not the right way to do this I think since I only want one set of values in this table, as I test this more i am getting overlapping data.

noplanman: So basically you want to update the values if the timestamp is the same, instead of creating a new entry every time?

Try with:
```php
$p = R::findOrCreate('monthlystats', ['date_checked' => $timestamp]);
```

db/monthly_stats.php
noplanman: Just leave it here. Sometimes SQL can be tricky, so it's totally fine to let PHP to some maths too :wink:

dmorley:
Well the current month does not work right, which means my math is off. If you divide out the current month from this test:

podupdb=# select to_char(date_checked, 'mm yy') AS yymm, sum(total_users) as users, count(DISTINCT to_char(date_checked, 'HH24 yy')), count(DISTINCT to_char(date_checked, 'dd yy')) from checks where online GROUP by yymm ORDER by yymm;
 yymm  |   users   | count | count 
-------+-----------+-------+-------
 01 18 | 480301944 |    24 |    31
 02 17 | 221407666 |    24 |    14
 02 18 | 445080768 |    24 |    28
 03 17 | 488240181 |    24 |    31
 03 18 | 398706179 |    24 |    26
 04 17 | 479812291 |    24 |    30
 04 18 |   2685717 |     4 |     4
 05 17 | 487784237 |    24 |    31
 06 17 | 462426452 |    24 |    30
 07 17 | 479846457 |    24 |    31
 08 17 | 621900775 |    24 |    31
 09 17 | 460900765 |    24 |    30
 10 17 | 476299967 |    24 |    31
 11 17 | 466428928 |    24 |    30
 12 17 | 489310309 |    24 |    31
(15 rows)

It's way off because the amount of checks. My brain says this should work and be right, it however is not

db/monthly_stats.php
```diff
@@ -0,0 +1,52 @@
...
+  $timestamp = $monthly['yymm'].'-01 01:01:01-01';
+  
+  try {
+    $p                   = R::findOrCreate('monthlystats', ['date_checked' => $timestamp]);
+    $p['total_users']    = $monthly['users'];
+    $p['total_posts']    = $monthly['posts'];
+    $p['total_comments'] = $monthly['comments'];
+    $p['total_pods']     = $monthly['pods'];
+    if($monthly['downtime']) {
+      $p['total_uptime']  =  round($monthly['uptime'] / $monthly['downtime'],0);
```
dmorley: Something not working right here, if you take the values manually from sql and do the maths it is right. Round here is giving wrong data it seems. So close but this last one not making sense

dmorley:
ok this is finally ready for a final review! @noplanman

db/pull.php
```diff
@@ -179,16 +179,11 @@
...
-  $ipv6        = false;
-  $iplookupv6  = explode(PHP_EOL, trim($delv->execute(['AAAA'], null, 15)->stdout));
-  $getaaaaonly = array_values(preg_grep('/\s+IN\s+AAAA\s+.*/', $iplookupv6));
-  if ($getaaaaonly) {
-    preg_match('/AAAA\s(.*)/', $getaaaaonly[0], $aaaaversion);
-    $ipv6 = trim($aaaaversion[1]) ?? '';
-  }
   $ip || $score -= 2;
 
+  $iplookupv6 = explode(PHP_EOL, trim($delv->execute(['AAAA'], null, 15)->stdout));
```
noplanman: @dmorley I've simplified this block, as the IPv6 address itself isn't used any more.

db/monthly_stats.php
```diff
@@ -0,0 +1,53 @@
...
+} catch (\RedBeanPHP\RedException $e) {
+  die('Error in SQL query: ' . $e->getMessage());
+}
+foreach ($monthly_totals as $monthly) {
+
+  // Format date to timestamp.
+  $timestamp = $monthly['yymm'].'-01 01:01:01-01';
+
+  try {
+    $p = R::findOrCreate('monthlystats', ['date_checked' => $timestamp]);
```
noplanman: If the row isn't found, this already initialises the new one with the `date_checked` field, so no need to set it further down (which is why I removed that line).

PR #138, opened 2018-04-19
Deadpods

dmorley:
admin can't delete anymore, the code should do its own cleanup

keep metadata if a pod dies rather than zero it out

hack for freindica getting de-listed due to odd dev code tags


go.php
```diff
@@ -27,7 +27,8 @@
       SELECT domain
       FROM pods
       WHERE signup
-        AND score > 90
+        AND uptime_alltime > 96
+        AND monthsmonitored > 2
```
noplanman: Yes! :+1:

db/pull.php
```diff
@@ -229,7 +232,7 @@
   _debug('Masterversion', $masterversion);
   $masterversioncheck = explode('.',$masterversion);
   $shortversioncheck = explode('.',$shortversion);
-  if (($masterversioncheck[1] - $shortversioncheck[1]) > 1) {
+  if (($masterversioncheck[1] - $shortversioncheck[1]) > 1 && strpos($shortversion,'dev') !== false) {
```
noplanman: Makes sense to add a short comment here to explain this "hack/fix".

showfull.php
```diff
@@ -1,14 +1,15 @@
...
+use CommerceGuys\Enum\AbstractEnum;
 
 defined('PODUPTIME') || die();
 
 try {
   $pods = R::getAll('
     SELECT domain, dnssec, podmin_statement, sslexpire, masterversion, shortversion, softwarename, monthsmonitored, score, signup, name, country, city, state, lat, long, uptime_alltime, active_users_halfyear, active_users_monthly, service_facebook, service_twitter, service_tumblr, service_wordpress, service_xmpp, latency, date_updated, ipv6, total_users, local_posts, comment_counts, userrating, status
     FROM pods
-    WHERE uptime_alltime > 50
+    WHERE status < 4
```
noplanman: Use the proper enum instead of an integer.
In this case `PodStatus::System_Deleted`

```php
  $pods = R::getAll('
    ....
    WHERE status < ?
    ...
  ', [PodStatus::System_Deleted]);
```

statsviewjs.php
```diff
@@ -12,7 +12,7 @@
       sum(total_users) AS users,
       round(avg(uptime_alltime),2) AS uptime
     FROM pods
-    WHERE NOT hidden
+    WHERE status < 4
```
noplanman: Use the proper enum instead of an integer.
In this case `PodStatus::System_Deleted`

```php
  $totals = R::getAll('
    ....
    WHERE status < ?
    ...
  ', [PodStatus::System_Deleted]);
```

showfull.php
```diff
@@ -1,14 +1,15 @@
 <?php
 
 use RedBeanPHP\R;
+use CommerceGuys\Enum\AbstractEnum;
```
noplanman: Not required here, can be removed.

podcrawler.php
```diff
@@ -14,16 +14,5 @@
 
 foreach($json->pods as $mydata) {
```
noplanman: Why not just use `$pods_json = file_get_contents($pods_json_url);` instead of all the `curl*` stuff to get the pods list?

podcrawler.php
```diff
@@ -14,16 +14,5 @@
...
-
-
-//local json from a diaspora pod json dump /admin_panel/pod/export
-$string = file_get_contents("pods.json");
-$json_a = json_decode($string, true);
-
-foreach ($json_a as $pod => $host) {
-  $addlocal = exec('php-cgi db/add.php domain='.$host['pod']['host']);
-  echo $addlocal;
+  echo $addjson."\r\n";
```
noplanman: Might as well make this a 1-liner:
```php
echo exec("php-cgi db/add.php domain={$mydata->host}") . "\r\n"; 
```

podcrawler.php
```diff
@@ -1,18 +1,6 @@
...
-curl_setopt($chss, CURLOPT_TIMEOUT, 19);
-curl_setopt($chss, CURLOPT_RETURNTRANSFER, 1);
-$output      = curl_exec($chss);
-curl_close($chss);
-$json = json_decode($output);
-
-foreach($json->pods as $mydata) {
-  $addjson = exec('php-cgi db/add.php domain='.$mydata->host);
-  echo $addjson."\r\n";
+$json = json_decode(file_get_contents('https://the-federation.info/pods.json'));
```
noplanman: For extra brownie points, pop this in an `if (...)` in case the list can't be loaded for some reason.
Otherwise `$json->pods` will explode.

Also, to be super safe and use PHP 7 features, prevent an explosion altogether with:
```php
foreach ($json->pods ?? [] as $poddata) {
```
This uses an empty array for the loop, if `$json->pods` doesn't exist or is set to `null`.

noplanman:
Oh, you merged already :grin: 
Well, then add that change to develop directly.

PR #139, opened 2018-05-01
cleanup and fix montly stats

dmorley:
monthly user stats were only good if you ran late in the day, clearly not right. and a few things for view-ability.


README.md
```diff
@@ -4,16 +4,7 @@ Poduptime is software to get live stats and data on listed Diaspora Pods.
...
-git
-curl
-postgresql postgresql-contrib
-wget
-dnsutils
-bind9
-npm
-nodejs nodejs-legacy
-composer
+php7.2 php7.2-curl php7.2-pgsql php-geoip php7.2-cli php7.2-common php7.2-json php7.2-readline php-cgi git curl postgresql postgresql-contrib wget dnsutils bind9 npm nodejs nodejs-legacy composer
```
noplanman: Yaay, PHP 7.2 :tada:

config.php.example
```diff
@@ -16,7 +16,7 @@ $pgpass = '';
 $pgdb = '';
 //admin email for forms
 $adminemail = '';
-//DNS server for dnssec testing
+//DNS server for dnssec testing. 1.1.1.1 tests the best
```
noplanman: Maybe use 1.1.1.1 as a fallback if no IP has been set here?

showfull.php
```diff
@@ -28,18 +28,18 @@
...
-    <th><a data-toggle="tooltip" data-placement="bottom" title="Does this pod allow new users.">Signups</a></th>
-    <th><a data-toggle="tooltip" data-placement="bottom" title="Number of total users on this pod.">Users</a></th>
+    <th><a data-toggle="tooltip" data-placement="bottom" title="Does this pod allow new users.">Signup</a></th>
+    <th><a data-toggle="tooltip" data-placement="bottom" title="Number of total users on this pod.">User</a></th>
     <th><a data-toggle="tooltip" data-placement="bottom" title="Number of users active last 6 months on this pod.">6m</a></th>
     <th><a data-toggle="tooltip" data-placement="bottom" title="Number of users active last 1 month on this pod.">1m</a></th>
-    <th><a data-toggle="tooltip" data-placement="bottom" title="Number of total posts on this pod.">Posts</a></th>
-    <th><a data-toggle="tooltip" data-placement="bottom" title="Number of total comments on this pod.">Comments</a></th>
-    <th><a data-toggle="tooltip" data-placement="bottom" title="How many months have we been watching this pod.">Months</a></th>
+    <th><a data-toggle="tooltip" data-placement="bottom" title="Number of total posts on this pod.">Post</a></th>
```
noplanman: Why singular instead of plural on all these?

showmap.php
```diff
@@ -69,11 +69,11 @@
       ?>
     ]
   };
-  var tiles = L.tileLayer('https://{s}.tiles.mapbox.com/v4/diasporg.l615e519/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoiZGlhc3BvcmciLCJhIjoibTVBaldtayJ9.HdGPBIFeZyNKKQqCmU11nA', {
+  var tiles = L.tileLayer('https://{s}.tiles.mapbox.com/v4/mapbox.streets/{z}/{x}/{y}.jpg70?access_token=pk.eyJ1IjoiZGlhc3BvcmciLCJhIjoibTVBaldtayJ9.HdGPBIFeZyNKKQqCmU11nA', {
```
noplanman: Should this access token be private, i.e. set in the config?
Or is it locked to domain name anyway?

showfull.php
dmorley: Since I added table borders things pushed out kinda far. Seemed like the font is already small and i did not want to make it any smaller, but needed just a hair more room for the table.

showmap.php
dmorley: Yes, I will fix.

Issue #158, opened 2018-06-18
redmatrix-> hubzilla


elvecio:
so now maybe we can change the name finally
https://github.com/diasporg/Poduptime/commit/aa99799eb3c0694d13420d72ea4e2630bdb92f6d#commitcomment-29394668

Issue #152, opened 2018-05-09
Security Check


dmorley:
db/pull.php is open to the webserver/public, should be cli only. If we want podmins to have an "update" button that should be after they get a key and do that from their podmin area. 

Are other files open that should not be?

Issue #143, opened 2018-05-09
Do version check better for pods


dmorley:
Dev pods do not really get a version check
Sometime prod code jumps from v 3.3 to 3.5 in a month and makes all pods stale
the checks table can be used to address this somehow, example:

SELECT DISTINCT ON (shortversion) shortversion, date_checked FROM checks WHERE domain = '' AND shortversion IS NOT NULL ORDER BY shortversion DESC LIMIT 2;

Should look to see what the dev norms are, how often etc. If a pod does not update for a year on dev what then? or prod ones that are a few versions behind but that's only a month etc. lots of fringe cases where code versions are not ==

dmorley:
Also all projects are on github and releases are available via the github API, that might be better then pulling random files from within repos to get version numbers off of

noplanman:
I think GitHub API is the way to go for the projects managed on GitHub.

https://developer.github.com/v3/repos/releases/

We'd have to make it flexible enough to allow variations for projects that might be managed elsewhere, like GitLab, BitBucket or even a completely different system.
Not sure how "in-the-future" this is though.

PR #159, opened 2018-06-22
use API for pulls of master version - enforce some security

dmorley:
Going to push this up for review, its been 4 weeks since I worked on it so it's getting stale.

https://github.com/diasporg/Poduptime/issues/152
https://github.com/diasporg/Poduptime/issues/143
https://github.com/diasporg/Poduptime/issues/158


noplanman:
Oooh, I'm liking this PR. Nice work!

Will review when I get a moment :+1:

db/podcrawler.php
```diff
@@ -1,12 +1,19 @@
...
-  $json = json_decode(file_get_contents('https://the-federation.info/pods.json'));
-  if ($json) {
-    foreach ($json->pods ?? [] as $poddata) {
-      echo exec("php-cgi add.php domain={$poddata->host}") . "\r\n";
-    }
+if (php_sapi_name() !== "cli") {
+  header('HTTP/1.0 403 Forbidden');exit;
+}
+
+$json = json_decode(file_get_contents('https://the-federation.info/graphql?query=%7B%0A%20%20nodes%20%7B%0A%20%20%20%20host%0A%20%20%7D%0A%7D&operationName=null&raw'));
```
noplanman: No need for pretty printing, I think it's better to keep this more readable:
`https://the-federation.info/graphql?query=%7Bnodes%7Bhost%7D%7D`

db/podcrawler.php
```diff
@@ -1,12 +1,19 @@
...
+if ($json) {
+  foreach ($json->data->nodes ?? [] as $poddata) {
+    echo exec("php-cgi add.php domain={$poddata->host}") . "\r\n";
   }
-} else {
-header('HTTP/1.0 403 Forbidden');
+}
+
+$json = json_decode(file_get_contents('https://diasp.org/pods.json'));
```
noplanman: Before adding these as a second step, I think it would be best to merge the-federation list and this list first and only run through the unique entries once. :+1:

db/pull-masterversions.php
```diff
@@ -14,32 +18,69 @@
 R::usePartialBeans(true);
 
 $softwares = [
```
noplanman: :tada: 

To make this more pretty, we'll OOPify this to make it more readable and understandable.
Mainly to prevent an epic if/else tree as more projects from different sources join the party.

db/pull.php
```diff
@@ -1,10 +1,17 @@
...
+  exit;
+}
+
 use RedBeanPHP\R;
 
-$debug   = isset($_GET['debug']) || (isset($argv) && in_array('debug', $argv, true));
-$newline = PHP_SAPI === 'cli' ? "\n" : '<br>';
+$debug    = isset($_GET['debug']) || (isset($argv) && in_array('debug', $argv, true));
+$sqldebug    = isset($_GET['sqldebug']) || (isset($argv) && in_array('sqldebug', $argv, true));
+$nowrite  = isset($_GET['nowrite']) || (isset($argv) && in_array('nowrite', $argv, true));
```
noplanman: Hmm, instead of saving the negative result, it's easier to go with the positive.

This makes if statements much easier to understand!

i.e.
```php
if ($write) {
  echo 'Write!';
}

// instead of
if (!$nowrite) {
  echo 'Write?';
}
```

db/gettoken.php
```diff
@@ -38,16 +38,16 @@
...
-  $output    = 'Link sent to your email';
+  $output    = 'Link sent to your email.';
 } elseif (!$pod['email']) {
-  die('domain is registered but no email associated, to add an email use the add a pod feature');
+  die('Domain is registered but no email associated, to add an email use the add a pod feature.');
 } else {
-  $to              = $adminemail;
-  $subject         = 'FORWARD REQUEST: Temporary edit key for ' . $_SERVER['HTTP_HOST'];
-  $message_lines[] = 'User trying to edit pod without email address.';
+  $to              = $pod['email'];
```
noplanman: :+1: Good call to not require admin intervention.

db/podcrawler.php
dmorley: Went a bit deeper as this thing takes forever right now

dmorley:
Ha, Learing Phpstorm and it looks like I somehow did work on my "wizard" branch on this branch, mess

noplanman:
Oh, could you maybe remove the wizard stuff for this PR?
Otherwise it will create a weird state where half a feature is added to a PR that is unrelated :grin: 

Also, I'll make a PR towards this PR with some other fixes I saw, instead of commenting them all here, ok?

dmorley:
Yup, manually cleaned it  up, I had just started on it. Sounds good to PR off my PR.

db/add.php
```diff
@@ -129,6 +133,6 @@
 
 } else {
   $log->lwrite('Could not validate your pod, check your setup! ' . $_domain);
-  echo 'Could not validate your pod, check your setup!<br>Take a look at <a href="https://' . $_domain . '/nodeinfo/2.0">your /nodeinfo</a>';
+  echo 'Could not validate your pod, check your setup!<br>Take a look at <a href="https://' . $link . '">your /nodeinfo</a>';
```
noplanman: You can remove the `https://`, as it's part of the `$link` already.

db/add.php
```diff
@@ -82,8 +82,12 @@
   }
 }
 
+if ($infos = json_decode(file_get_contents('https://' . $_domain . '/.well-known/nodeinfo'), true)) {
+  $link = end($infos['links'])['href'];
```
noplanman: If there are multiple links, it's always the last one?

dmorley: I think so, however I could not find anyone with > 2 links.
The spec seems to put them in order so I feel like the answer is yes 3.0 will be last and end should be safe. Think I need to pull the numbers off the end and use that?
https://github.com/jhass/nodeinfo/blob/master/PROTOCOL.md

Issue #147, opened 2018-05-09
visual table cleanup


dmorley:
filters - the detailed view is crazy long and wide
pagination - the detailed view is crazy long and wide

PR #161, opened 2018-06-26
Add filter, paging, update menus to make them mobile

dmorley:
I was going to go back to wizard, but I realized it want the wizard to adjust the filter real time in the background as it goes, but no filter. 
Added filter, paging. Updated tablesort for these new feaures, updated bootstrap to make that work, updated the menus since bootstrap is not the same. 
400% improvement on mobile.

https://github.com/diasporg/Poduptime/issues/147


noplanman:
Have you seen this?
https://bower.io/blog/2017/how-to-migrate-away-from-bower/

I haven't looked into it, just got that popup when using `yarn install` for the first time.

dmorley:
Yea that's what prompted my other pr moving from bower to yarn

Issue #142, opened 2018-05-09
update bootstrap and remove old facebox


dmorley:
facebox.js is way old and ugly. bootstrap can do this better probably

dmorley:
Add https://yarnpkg.com/en/package/@fancyapps/fancybox to yarn
Remove bower.json and bower stuff after and only use yarn

PR #162, opened 2018-06-26
bower to yarn

dmorley:
bower complaining it is dead. Yarn seems legit?


noplanman:
Can `bower.json` be removed completely?

composer.json
```diff
@@ -1,9 +1,7 @@
 {
     "require": {
         "noplanman/xec": "0.1.0",
-        "gabordemooij/redbean": "^5.0",
```
noplanman: :exclamation: These are still required for PHP

dmorley:
facebox does not exist elsewhere and removing it is a project, bootstrap modals do not work with graph.js etc. Maybe I should just add a copy of facebox.js to /js for now since it won't ever get an update?

composer.json
dmorley: Yes I left these two in composer.json but crawler and enum are in yarn now. That works?

noplanman: yarn -> Javascript dependencies
composer -> PHP dependencies

So the enum and crawler you've added with yarn can be removed again, as they aren't being used via JS.

dmorley: ohhh!

package.json
```diff
@@ -7,10 +7,7 @@
     "bootstrap": "^4.1.1",
     "bower-resolve": "^2.2.1",
     "chart.js": "^2.7.2",
-    "crawler-detect": "^1.0.1",
-    "enum": "^2.5.0",
     "jquery": "^3.3.1",
-    "jquery-ui": "^1.12.1",
```
dmorley: this is unused as far as I can see anywhere

noplanman:
Hmm, i think we find an alternative and replace facebox?
Last commit is from 2013 :confused:
https://github.com/defunkt/facebox

Maybe fancybox?
https://yarnpkg.com/en/package/@fancyapps/fancybox

dmorley:
Looks nice. It's a bit messy but let's do that as part of https://github.com/diasporg/Poduptime/issues/142 and make a note there that bower.json then goes away.

Issue #144, opened 2018-05-09
language and location for go.php


dmorley:
Should we be thinking about language and location on go.php redirects?

noplanman:
I guess this kind of ties in with the wizard (#145), the only problem here, is getting the prevalent language of the pod.

Because really that's what counts, right?
Even if the pod is in an English speaking country for example, if all users speak German, the language the new user would choose should be German I think.

dmorley:
Going to close this and let the wizard or other things use the autolanguage as desired.

PR #165, opened 2018-07-01
detect language from body text of main page

dmorley:
https://github.com/diasporg/Poduptime/issues/144

Pull language off the web page. It tests pretty decent. Many non-en pods have en splash pages because that is the default in the source code.


noplanman:
Have merged develop into this PR :+1:

Issue #146, opened 2018-05-09
Userratings cleanup


dmorley:
Make popup's look nice and match the site.
Don't set 10 as a default
Make it so a podmin can link to site with their pod rating form shown
Show ratings on simple view somehow
Deal with a pod with more than a few ratings and how that will show

dmorley:
Leaving open for the task of making it linkable that needs to wait for new popup js or something first

PR #167, opened 2018-07-03
restore user ratings, tiny clean up on looks

dmorley:
https://github.com/diasporg/Poduptime/issues/146

Light clean up here, it's actually a bug since we moved to redbean since they do not support a table_name only tablename. no wonder no one uses!

This will get cleaned more in the bootstrap, facebox and tables clean up also so I left those items alone for now and kept it to just fixing broken and make it slightly less ugly and work properly

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+# 2x+
+
+## DB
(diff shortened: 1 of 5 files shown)


noplanman:
Have merged develop into this PR :+1:

PR #174, opened 2018-07-06
Facebox to featherlight js for lightbox

dmorley:
https://github.com/diasporg/Poduptime/issues/142

Also split out the charts so it's not a mess on a mobile device. I only linked once, I could of course link to every user/comment count etc, seems overkill, not sure.

PR to the devcleanup to keep simple to merge

Added popper.js as it is required by bootstrap 4.1.1


showfull.php
```diff
@@ -64,7 +64,7 @@
     <tbody>
     <?php
     foreach ($pods as $pod) {
-        $pod_name       = htmlentities($pod['name'] ?? '', ENT_QUOTES);
+        $pod_name       = ($pod_name ? htmlentities($pod['name'], ENT_QUOTES) : '');
```
noplanman: `$pod_name` doesn't exist here.

So I assume you mean to keep this?
```php
$pod_name = htmlentities($pod['name'] ?? '', ENT_QUOTES);
```

Also, this value isn't used anywhere, so maybe just remove this line altogether?
Same thing for `show.php`.

showfull.php
```diff
@@ -121,7 +121,7 @@
         echo '</td>';
 
         $podmin_statement = htmlentities($pod['podmin_statement'] ?? '', ENT_QUOTES);
-        echo '<td data-text="' . $podmin_statement . '">' . ($podmin_statement ? '<a tabindex="0" data-toggle="tooltip" data-trigger="focus" data-placement="left" data-html="true" title="' . $podmin_statement . '">&#128172;</a>' : '&nbsp;') . '</td></tr>';
+        echo '<td>' . ($podmin_statement ? '<a href="#" data-featherlight="<p>' . $podmin_statement . '</p>">&#128172;</a>' : '&nbsp;') . '</td></tr>';
```
noplanman: :+1: 

Much better to have a dedicated modal popup than the tooltip, nice!

noplanman:
This PR is looking good, me gusta

Issue #149, opened 2018-05-09
Cookie me


dmorley:
Maybe once you pick a pod we help you remember it via a cookie? 
I notice a lot of new users sign up and then do not remember the xxx.xxx domain of where they signed up. Maybe if we store that and know that at least we could suggest they go back there? 
Are cookies still bad? Or good?

noplanman:
As we'd literally be using them "to improve the user experience", I think it would be totally fine.

The only question I have here, is how we know which pod the user registers on.
Say a user opens multiple pods directly from the list (which all go through `go.php`...), we won't be able to know which one the user chose in the end.
Same thing applies for the "Auto choose", which could be clicked but then the opened pod isn't the one the user wants to register on.

If we want to add this, we'd have to either save all the clicked links, or just the last one I guess :thinking: 
What do you think?

dmorley:
I think the last. See my comment on dia.so also on other issue. Maybe that's a better place or moving that dia.so to podupti.me somehow. Or I am just making things more complex trying to wrap that into here!

PR #177, opened 2018-07-07
Simple cookie hint

dmorley:
https://github.com/diasporg/Poduptime/issues/149

Very basic to try and help users remember the pod they picked (last in the simple case)

### DIFF
--- a/go.php
+++ b/go.php
@@ -48,7 +48,7 @@ try {
     if (!(new CrawlerDetect())->isCrawler()) {
         R::store($c);
     }
-    setcookie("domain", $domain, 2147483647);
+    setcookie("domain", $domain, Carbon::now()->addCentury()->timestamp);
     header('Location: https://' . $domain);
 } catch (\RedBeanPHP\RedException $e) {
     die('Error in SQL query: ' . $e->getMessage());
--- a/go.php
+++ b/go.php
@@ -48,7 +48,7 @@ try {
     if (!(new CrawlerDetect())->isCrawler()) {
         R::store($c);
     }
-
+    setcookie("domain", $domain, 2147483647);
     header('Location: https://' . $domain);
 } catch (\RedBeanPHP\RedException $e) {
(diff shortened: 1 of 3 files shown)


go.php
```diff
@@ -48,7 +48,7 @@
     if (!(new CrawlerDetect())->isCrawler()) {
         R::store($c);
     }
-
+    setcookie("domain", $domain, 2147483647);
```
noplanman: Erm, expiry should be in relation to now, no?
As in `Carbon::now()->addCentury()->timestamp;`

show.php
```diff
@@ -47,8 +47,11 @@
         $pod_name       = htmlentities($pod['name'], ENT_QUOTES);
         $humanmonitored = Carbon::now()->subDays($pod['daysmonitored'])->diffForHumans(null, true);
         $tip            = "This {$pod['softwarename']} pod's uptime is {$pod['uptime_alltime']}% over {$humanmonitored}.";
-        echo '<tr><td><div title="' . $tip . '" data-toggle="tooltip" data-placement="bottom"><a class="text-success url" target="_self" href="/go.php?domain=' . $pod['domain'] . '">' . $pod['domain'] . '</a></div></td>';
-
+        if ($_COOKIE["domain"] === $pod['domain']) {
```
noplanman: If the cookie hasn't been set yet, this throws an error, so fix it with:
```php
...
if ($_COOKIE['domain'] ?? null === $pod['domain']) {
...
```

:exclamation: Same applies to `showfull.php`

noplanman:
Might be good to have a popup asking about saving the cookie?

Hmm, or maybe because this cookie is actually a true helper, just let it do it's job :wink:

dmorley:
Let's not ask, but we could put one of them cookie things saying we use them? the sorter also sets several cookies now to remember your sort

noplanman:
> we could put one of them cookie things saying we use them?

Yep :+1:

PR #179, opened 2018-07-08
move to supported db for IP geo data

dmorley:


config.php.example
```diff
@@ -5,13 +5,13 @@
 
 declare(strict_types=1);
 
-//backup directory
+//backup directory - app directory path
```
noplanman: I think it would make sense to have these setting all as full paths, then just use `__DIR__ . '/whatever'` as the default. Otherwise it's confusing, especially when you want to set different paths that are outside the project directory, which really should be the case for logs and backups!

db/pull.php
```diff
@@ -235,14 +239,15 @@
...
-    $location = geoip_record_by_name($ip);
-    _debug('Location', $location, true);
-    $countryname = !empty($location['country_name']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['country_name']) : null;
-    $country     = !empty($location['country_code']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['country_code']) : null;
-    $city        = !empty($location['city']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['city']) : null;
-    $state       = !empty($location['region']) ? iconv('UTF-8', 'UTF-8//IGNORE', $location['region']) : null;
-    $lat         = !empty($location['latitude']) ? $location['latitude'] : 0;
-    $long        = !empty($location['longitude']) ? $location['longitude'] : 0;
+    $geo = $reader->city($ip);
+    $countryname = !empty($geo->country->name) ? iconv('UTF-8', 'UTF-8//IGNORE', $geo->country->name) : null;
```
noplanman: Are all these `iconv` still necessary with this library?

Also, to make it PHP 7 friendly and shorter:
```php
$countryname = iconv('UTF-8', 'UTF-8//IGNORE', $geo->country->name ?? null) ?: null;
...
// Lat/Long
$lat         = $geo->location->longitude ?? null ?: 0;
```

db/pull.php
```diff
@@ -240,12 +240,12 @@
     _debug('Iplookupv6', $iplookupv6, true);
 
     $geo = $reader->city($ip);
-    $countryname = !empty($geo->country->name) ? iconv('UTF-8', 'UTF-8//IGNORE', $geo->country->name) : null;
-    $country     = !empty($geo->country->isoCode) ? iconv('UTF-8', 'UTF-8//IGNORE', $geo->country->isoCode) : null;
-    $city        = !empty($geo->city->name) ? iconv('UTF-8', 'UTF-8//IGNORE', $geo->city->name) : null;
-    $state       = !empty($geo->mostSpecificSubdivision->name) ? iconv('UTF-8', 'UTF-8//IGNORE', $geo->mostSpecificSubdivision->name) : null;
-    $lat         = !empty($geo->location->latitude) ? $geo->location->latitude : 0;
-    $long        = !empty($geo->location->longitude) ? $geo->location->longitude : 0;
+    $countryname = $geo->country->name ?? null ?: 0;
```
noplanman: These should probably be `null` if no value exists instead of `0`.
Otherwise it will show in the list as `City 0` (or something like that)

Maybe `lat` and `long` as well?

db/pull.sh
```diff
@@ -43,11 +44,19 @@ if [ "$HOUR" = 23 ]; then
...
     echo "$SAD"
   fi
-  printf "%s" "Updating CA..."
-  if wget -q https://curl.haxx.se/ca/cacert.pem -O ../cacert.pem; then
-    echo "$HAPPY"
-  else
-    echo "$SAD"
+  if [ "$DAY" = 23 ]  || [ "$@" = "init" ]; then
+    printf "%s" "Updating CA..."
+    if wget -q https://curl.haxx.se/ca/cacert.pem -O ../cacert.pem; then
```
noplanman: Maybe also use `curl` here to not require `wget`?

(Yep, we're entering micro-optimisation territory :joy_cat:)

config.php.example
```diff
@@ -5,10 +5,10 @@
 
 declare(strict_types=1);
 
-//backup directory - app directory path
+//backup directory - full dir path
 $backup_dir = '/backup';
```
noplanman: These defaults should be changed to `__DIR__ . '/backup'` etc.

dmorley: I don't fully understand the magic of __DIR__

noplanman: `__DIR__` returns the absolute path of the current file.
As easy as that :+1: 

You can test is with this:
`php -r 'echo __DIR__;'`

config.php.example
```diff
@@ -32,8 +32,11 @@ $adminemail = '';
 //DNS server for dnssec testing. 1.1.1.1 tests the best
 $dnsserver = '';
 
-//CA for curl to use - suggest wget http://curl.haxx.se/ca/cacert.pem as they pull from mozilla if you use a system CA then certs like wosign and startssl will be valid but users will be unable to connect to them
+//CA for curl to use - full file path (pull.sh will update monthly to this directory as cacert.pem)
```
noplanman: This would also be the `dir path`, not the `file path` (i.e. `/path` vs `/path/cacert.pem`)

Also, I noticed that these paths aren't actually being used in the shell script!
(see comments below regarding the shell script)

db/pull.sh
noplanman: To get the path from `config.php`:
```bash
...
CACERT_FILE="$(php -r "include __DIR__ . '/../config.php'; echo \$cafullpath;")"
if wget -q https://curl.haxx.se/ca/cacert.pem -O "$CACERT_FILE"; then
...
```

Same procedure for GeoLite DB file below.

PR #180, opened 2018-07-09
start testing testing

dmorley:


dmorley:
Maybe we test new things vs go back to old ones? Not the best but it's something and I have no idea how to write a test, good to learn even if our coverage is .001%

dmorley:
Them die's kill lots of things, who the heck codes like that  :neckbeard:  :scream_cat:

yarn.lock
```diff
@@ -0,0 +1,66 @@
+# yarn lockfile v1
```
noplanman: I guess this guy shouldn't be in this PR :confused: 

Best to just add it directly to the `develop` branch

dmorley: I added it cuz the travis directions say if it sees yarn.lock it will install and run yarn. Does that make any sense?

noplanman: Ah, ok, didn't know that!

I'd normally run all commands in travis manually (i.e. via `travis.yml`) instead of relying on auto-magic that might change.

dmorley:
Actually let's put this in the branch, you can't test travis and codacy without the files being in the release, it will be a failing disaster for now but at least it's going to start to CI things, it will not even look since master is the only branch included

noplanman:
I think we start with the testing things after the next release, as not to overdo things.
For the tests, a lot of the code needs to be changed, so best draw a line and then take the next plunge :+1:

PR #190, opened 2018-07-18
Ui easy mods

noplanman:
Make table headers DIV instead of A tags.
Add tooltips to services images.

The weird tooltip flashing is still an issue when only few results are displayed.
No idea how to fix that :confused:


dmorley:
flashytips lets call them. a fight for another release

